Issue #20, opened 2026-04-03
Start Architecture Decision Records (ADRs)


ojongerius:
## Problem

Protocol and SDK design decisions are undocumented. New contributors and AI agents see *what* was chosen but not *why*. This leads to well-intentioned changes that accidentally violate design constraints.

## Proposal

Create a `docs/adr/` directory with lightweight decision logs using the Nygard template:

```markdown
# ADR-NNNN: Title

## Status
Proposed | Accepted | Deprecated | Superseded by ADR-XXXX

## Context
What is the issue that motivates this decision?

## Decision
What is the change that we're proposing?

## Consequences
What becomes easier or more difficult because of this change?
```

Good initial ADRs for this project:

- Why Ed25519 (not ECDSA or RSA) for receipt signing
- Why RFC 8785 for JSON canonicalization
- Why SQLite for local receipt storage
- Why three separate SDK implementations (not code generation from spec)
- Why YAML for the policy engine (mcp-proxy)
- Why W3C Verifiable Credentials as the receipt envelope

## References

- [ADR GitHub org](https://adr.github.io/)
- [AWS ADR best practices](https://aws.amazon.com/blogs/architecture/master-architecture-decision-records-adrs-best-practices-for-effective-decision-making/)
- [adr-tools CLI](https://github.com/npryce/adr-tools)

PR #79, opened 2026-04-05
docs: ADR-0002 RFC 8785 for JSON canonicalization

### LINKED ISSUE #40: ADR: Why RFC 8785 for JSON canonicalization
Write an ADR documenting why RFC 8785 (JSON Canonicalization Scheme) was chosen for deterministic serialization before signing and hashing.

Follow the template in `docs/adr/0000-template.md`.

Parent issue: #20

ojongerius:
## Summary

- Add ADR-0002 documenting the decision to use RFC 8785 (JSON Canonicalization Scheme) for canonical JSON serialization in receipt signing and chain hashing
- Covers alternatives considered (JSON-LD canonicalization, custom canonical form, binary formats), security considerations for cross-implementation divergence, and test vector requirements
- Update ADR index in docs/adr/README.md

Closes #40
Refs #20

## Issues raised during review

- #82 — Go SDK `sort.Strings` does not implement RFC 8785 UTF-16 code unit ordering
- #84 — Spec uses `validFrom` but all SDKs use `issuanceDate`
- #85 — Cross-SDK nullable field handling divergence risk
- #86 — Python SDK `_utf16_sort_key` uses UTF-16-LE bytes, not UTF-16 code units

## Test plan

- [x] Verify ADR content is accurate against spec section 7.1 and SDK implementations
- [x] Confirm ADR index in README.md renders correctly


docs/adr/0002-rfc8785-json-canonicalization.md
```diff
@@ -0,0 +1,86 @@
...
+- **Proven in adjacent ecosystems** — RFC 8785 is used in JWS (JSON Web Signature) and JWK (JSON Web Key) contexts, and is referenced by the W3C VC Data Integrity specification as an alternative to JSON-LD canonicalization.
+- **Alignment with W3C VC Data Integrity** — the spec notes (§7.1) that the canonicalization step is aligned with RFC 8785 usage in the W3C VC Data Integrity specification, though the overall signing procedure is intentionally simplified.
+
+## Security Considerations
+
+### Cross-implementation divergence
+
+The security of every receipt signature depends on all SDK implementations producing byte-identical canonical output for the same input. If two implementations canonicalize a receipt differently, signatures created by one will not verify under the other. This is a real attack surface: an attacker who can construct a receipt that canonicalizes differently across SDKs could present a receipt that appears valid to one verifier but invalid to another, or produce two different hashes for the same logical receipt.
+
```
ojongerius: **Precision: TypeScript doesn't "implement RFC 8785 directly"**

The ADR states all three SDKs "implement it directly without external dependencies." This is accurate for Go and Python, but the TypeScript SDK's approach is qualitatively different. Looking at `sdk/ts/src/receipt/hash.ts`:

- **Key sorting** uses `Object.keys(value).sort()`, which works because JS strings are UTF-16 and `.sort()` does lexicographic comparison by UTF-16 code units — this *coincidentally matches* RFC 8785, but isn't an explicit implementation of the spec's sorting rule.
- **Number serialization** uses `String(n)` (i.e., ES6 `Number.prototype.toString()`), which is exactly what RFC 8785 mandates — so the TS SDK *delegates to the runtime* rather than re-implementing the algorithm.

The current wording "all three SDKs implement it directly" implies the same kind of from-scratch implementation in each. A more precise description would be something like: "The TypeScript SDK relies on JavaScript's native UTF-16 string comparison and ES6 number serialization, which inherently conform to RFC 8785; Go and Python re-implement these algorithms explicitly."

This matters because the TS SDK's correctness is a property of the JS runtime, not of explicit code — which is actually a *strength* (less room for bugs), but the ADR should be clear about it.

docs/adr/0002-rfc8785-json-canonicalization.md
```diff
@@ -0,0 +1,86 @@
...
+This risk is the same canonicalization dependency identified in ADR-0001's Security Considerations. Mitigation requires:
+
+1. Shared cross-language test vectors (`sdk/py/tests/fixtures/ts_vectors.json`) that all SDKs must pass.
+2. Test coverage for edge cases: Unicode strings, negative zero, deeply nested objects, empty objects and arrays, and keys that sort differently under UTF-8 vs UTF-16 ordering.
```
ojongerius: **Missing security consideration: Python's \`exclude_none\` and nullable field preservation**

The Python SDK uses \`model_dump(by_alias=True, exclude_none=True)\` in all canonicalization paths (hash, sign, verify). This silently drops any field with a \`None\` value from the canonical representation. The SDK then has special re-insertion logic for \`previous_receipt_hash\` to preserve it as explicit \`null\`:

\`\`\`python
# sdk/py/src/agent_receipts/receipt/hash.py:97-101
cs = d.get("credentialSubject", {})
chain = cs.get("chain", {})
if "previous_receipt_hash" not in chain:
    chain["previous_receipt_hash"] = None
\`\`\`

This pattern has two risks worth documenting:

1. **Any new nullable field** added to the receipt schema must be audited for whether \`None\` is semantically meaningful (like \`previous_receipt_hash\` for the first receipt in a chain). If it is, it needs the same re-insertion treatment, or Python will produce different canonical bytes than Go/TypeScript.

2. **Go and TypeScript handle this differently** — Go uses \`omitempty\` JSON tags and TypeScript serializes all fields. If a field is \`None\`/\`null\`/zero-value, the three SDKs may disagree on whether to include it, which is exactly the kind of canonicalization divergence this ADR's security section warns about.

This is a concrete instance of the "cross-implementation divergence" risk class and deserves a mention in the Security Considerations section.

docs/adr/0002-rfc8785-json-canonicalization.md
```diff
@@ -0,0 +1,91 @@
...
+
+Any implementation that fails to exclude `proof` before canonicalization will produce signatures and hashes that are incompatible with all other implementations. This invariant must be enforced by cross-language test vectors.
+
+### Unicode edge cases
+
+RFC 8785 specifies that object keys are sorted by UTF-16 code unit order, not by Unicode code point order. For characters in the Basic Multilingual Plane (U+0000 to U+FFFF), these orderings are identical. They diverge for supplementary characters (emoji, CJK Extension B, etc.) which are represented as surrogate pairs in UTF-16. Implementations in languages that use UTF-8 natively (Go, Python) must correctly implement UTF-16 code unit comparison for key sorting. Python does this via `_utf16_sort_key()` which encodes keys as UTF-16-LE for comparison. The Go SDK currently uses `sort.Strings()` which sorts by UTF-8 byte order — this is correct for BMP characters but incorrect for supplementary characters (#82). The current SDKs include Unicode string test cases, but supplementary-character key ordering should be explicitly covered.
```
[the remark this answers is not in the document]
ojongerius: Good catch on both points.

1) You're right — UTF-16-LE byte comparison is not equivalent to UTF-16 code unit comparison. Little-endian byte order means the low byte comes first, so `U+00FF` ([0xFF, 0x00]) sorts *after* `U+0100` ([0x00, 0x01]) in byte order, but *before* it in code unit order. The fix is to use UTF-16-BE (or compare 16-bit integers). Filed #86.

2) Also correct that the shared cross-language vectors only use ASCII keys. Will fix the wording.

PR #167, opened 2026-04-14
fix(mcp-proxy): default approval HTTP server to random port

ojongerius:
## Summary

- Default `-http` to `127.0.0.1:0` so the OS assigns a free port instead of crashing the second instance with `bind: address already in use`. This fixes the case where Codex + Claude Code (or two Claude sessions) run simultaneously and the client sees `connection closed: initialize response`.
- Log the bound address + token to stderr in two formats: a human-readable copy-pasteable line, and a JSON event line as a minimal discovery primitive for future approval tooling.
- Drop the now-obsolete port-conflict workaround from the Claude Code and Codex gotcha sections; replace with a one-line note. Users who want a stable URL can still pin with `-http 127.0.0.1:PORT`.

## Test plan

- [x] `go build ./... && go vet ./...` clean
- [x] `go test ./...` passes
- [x] Handshake works with default (random port) while another `mcp-proxy` already owns 8080
- [x] Handshake works with explicit `-http 127.0.0.1:18099` (escape hatch)
- [x] Explicit pin to a busy port still fails fast with `log.Fatalf` (correct behaviour — user explicitly asked for that port)
- [x] `8080` references audited; remaining ones are intentional ("pin to 8080 if you want stability" examples) plus the unrelated dashboard binary

### DIFF
--- a/mcp-proxy/README.md
+++ b/mcp-proxy/README.md
@@ -112,11 +112,13 @@ rules:
 
 Actions: `pass` (log only), `flag` (log + highlight), `pause` (wait for approval), `block` (reject).
 
-When a tool call is paused, approve or deny via HTTP:
+When a tool call is paused, approve or deny via HTTP. The approval URL and bearer token are logged to stderr at startup (the default port is random; pass `-http 127.0.0.1:PORT` to pin):
 
 ```sh
-curl -X POST http://localhost:8080/api/tool-calls/{id}/approve
-curl -X POST http://localhost:8080/api/tool-calls/{id}/deny
+curl -X POST http://127.0.0.1:PORT/api/tool-calls/{id}/approve \
+  -H "Authorization: Bearer $APPROVAL_TOKEN"
+curl -X POST http://127.0.0.1:PORT/api/tool-calls/{id}/deny \
+  -H "Authorization: Bearer $APPROVAL_TOKEN"
 ```
 
 Paused calls auto-deny after 60 seconds (fail-safe).
--- a/mcp-proxy/README.md
+++ b/mcp-proxy/README.md
@@ -112,9 +112,11 @@ rules:
 
 Actions: `pass` (log only), `flag` (log + highlight), `pause` (wait for approval), `block` (reject).
 
-When a tool call is paused, approve or deny via HTTP. The approval URL and bearer token are logged to stderr at startup (the default port is random; pass `-http 127.0.0.1:PORT` to pin):
+When a tool call is paused, approve or deny via HTTP. The approval URL and bearer token are logged to stderr at startup (the default port is random; pass `-http 127.0.0.1:PORT` to pin). Copy the token from the startup line and export it before running the curls:
 
 ```sh
+export APPROVAL_TOKEN=<token-from-stderr>
+
 curl -X POST http://127.0.0.1:PORT/api/tool-calls/{id}/approve \
   -H "Authorization: Bearer $APPROVAL_TOKEN"
 curl -X POST http://127.0.0.1:PORT/api/tool-calls/{id}/deny \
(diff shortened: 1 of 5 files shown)


ojongerius (commented):
Thanks for the review! Addressed all four comments in 98c5679. Per-comment responses inline.

mcp-proxy/cmd/mcp-proxy/main.go
```diff
@@ -227,6 +227,16 @@ func serve() {
 		if err != nil {
 			log.Fatalf("mcp-proxy: http server: %v", err)
 		}
+		approvalURL := "http://" + ln.Addr().String()
+		// Human-readable line (one copy-pasteable string).
+		log.Printf("mcp-proxy: approvals at %s (token: %s)", approvalURL, approvalToken)
```
ojongerius, line 232: Good catch — fixed in 98c5679. Switched the human-readable line to `fmt.Fprintf(os.Stderr, ...)` so it has no timestamp prefix and matches the sample in the docs verbatim.

mcp-proxy/cmd/mcp-proxy/main.go
```diff
@@ -227,6 +227,16 @@ func serve() {
...
+		// Machine-readable line — minimal discovery primitive for future tooling.
+		endpointJSON, _ := json.Marshal(map[string]string{
+			"event": "approval_endpoint",
+			"url":   approvalURL,
+			"token": approvalToken,
+		})
```
ojongerius, line 238: Fair — fixed in 98c5679. Now logs the marshal error and falls back to `{"event":"approval_endpoint"}` so any future tooling that parses the discovery line still sees a valid event name. Agreed it's unlikely to fail in practice for a static string map, but the cost of handling it is trivial and it keeps the discovery contract honest.

site/src/content/docs/mcp-proxy/configuration.mdx
```diff
@@ -118,21 +118,31 @@ A bundled [`configs/github_taxonomy.json`](https://github.com/agent-receipts/ar/
...
 ```bash
 # Approve
-curl -X POST http://localhost:8080/api/tool-calls/{id}/approve \
+curl -X POST http://127.0.0.1:59850/api/tool-calls/{id}/approve \
   -H "Authorization: Bearer $APPROVAL_TOKEN"
 
 # Deny
-curl -X POST http://localhost:8080/api/tool-calls/{id}/deny \
+curl -X POST http://127.0.0.1:59850/api/tool-calls/{id}/deny \
   -H "Authorization: Bearer $APPROVAL_TOKEN"
```
ojongerius, line 139: Good point — fixed in 98c5679. Added an `export APPROVAL_TOKEN=5fce4e79...` line with a comment telling the reader to copy the token from the stderr line above (or parse the JSON event), so `$APPROVAL_TOKEN` in the curl examples is no longer dangling.

mcp-proxy/README.md
```diff
@@ -112,11 +112,13 @@ rules:
...
-When a tool call is paused, approve or deny via HTTP:
+When a tool call is paused, approve or deny via HTTP. The approval URL and bearer token are logged to stderr at startup (the default port is random; pass `-http 127.0.0.1:PORT` to pin):
 
 ```sh
-curl -X POST http://localhost:8080/api/tool-calls/{id}/approve
-curl -X POST http://localhost:8080/api/tool-calls/{id}/deny
+curl -X POST http://127.0.0.1:PORT/api/tool-calls/{id}/approve \
+  -H "Authorization: Bearer $APPROVAL_TOKEN"
+curl -X POST http://127.0.0.1:PORT/api/tool-calls/{id}/deny \
+  -H "Authorization: Bearer $APPROVAL_TOKEN"
```
ojongerius, line 121: Same fix here — 98c5679 adds an `export APPROVAL_TOKEN=<token-from-stderr>` line before the curl snippet, and the surrounding prose points readers at the startup stderr line as the source of the token.

PR #178, opened 2026-04-15
fix(sdk-py): resolve all 27 pyright strict errors and gate on CI

### LINKED ISSUE #175: sdk/py: fix pyright strict errors and gate on CI
## Problem

`sdk/py/CLAUDE.md` tells contributors that **168 pyright errors are "pre-existing and expected"**. Actual count verified today: **27 errors, not 168** — the doc is badly stale. Worse, the framing teaches agents to ignore the type-check signal entirely.

Pyright is declared in `pyproject.toml:39` (`pyright>=1.1`) with `typeCheckingMode = "strict"` (`pyproject.toml:64-67`) but is not run in CI (`.github/workflows/sdk-py.yml` runs ruff and pytest only). So type errors accumulate unchecked.

## What pyright actually reports

27 errors across 5 files (`cd sdk/py && uv run pyright src`):

| File | Count | Nature |
|---|---|---|
| `src/agent_receipts/receipt/hash.py` | 11 | `canonicalize(value: Any)` + `hash_receipt(receipt: Any)` cascade unknown types through recursion |
| `src/agent_receipts/receipt/create.py` | 7 | All on line 92 — uses `UnsignedAgentReceipt(**{"@context": ..., ...})` dict-spread so pyright can't see individual kwargs. Fix: switch to direct kwargs (the model has `populate_by_name=True` at `types.py:144`, so `context=list(CONTEXT), id=..., ...` works). |
| `src/agent_receipts/taxonomy/config.py` | 7 | Untyped JSON/YAML loads; narrow with TypedDict after `json.loads` |
| `src/agent_receipts/receipt/signing.py` | 1 | Redundant `isinstance(x, str)` at line 106 |
| `src/agent_receipts/receipt/chain.py` | 1 | Loose `receipts` parameter typing |

**Error categories:** 11 × `reportUnknownVariableType`, 7 × `reportArgumentType`, 6 × `reportUnknownMemberType`, 2 × `reportUnknownArgumentType`, 1 × `reportUnnecessaryIsInstance`.

Most are straightforward. The `hash.py` cluster is the only one needing real thought — either tighten recursion type signatures (preferred) or use explicit `cast(...)` at the boundary between `Any`-typed input and typed inner logic.

## Scope

1. Fix the 27 errors.
2. Add `uv run pyright src` as a CI step in `.github/workflows/sdk-py.yml` so regressions are caught.
3. Delete the "168 pre-existing Pydantic errors are expected" line from `sdk/py/CLAUDE.md`.

## Non-goals

- Do not relax `typeCheckingMode` from `strict` — the goal is to meet the bar the project has already declared.
- Do not add Pydantic-specific pyright plugin configuration; it's not the root cause.

## Acceptance

- `uv run pyright src` exits 0 on all Python versions in the CI matrix (3.11, 3.12, 3.13).
- CI fails on new pyright regressions.
- `sdk/py/CLAUDE.md` no longer claims 168 errors are expected.

---
_Source: internal code-review pass, April 2026._

ojongerius:
Closes #175.

### DIFF
--- a/.github/workflows/sdk-py.yml
+++ b/.github/workflows/sdk-py.yml
@@ -37,4 +37,6 @@ jobs:
 
       - run: uv run ruff format --check .
 
+      - run: uv run pyright src
+
       - run: uv run pytest -v
--- a/sdk/py/CLAUDE.md
+++ b/sdk/py/CLAUDE.md
@@ -11,7 +11,7 @@ uv sync --all-extras       # install deps
 uv run pytest -v           # run tests
 uv run ruff check .        # lint
 uv run ruff format .       # format
-uv run pyright src         # type check (168 pre-existing Pydantic errors are expected)
+uv run pyright src         # type check (strict mode, must pass)
 ```
 
 ## Architecture
--- a/sdk/py/src/agent_receipts/receipt/chain.py
+++ b/sdk/py/src/agent_receipts/receipt/chain.py
@@ -29,7 +29,9 @@ class ChainVerification:
 
     valid: bool
     length: int
-    receipts: list[ReceiptVerification] = field(default_factory=list)
+    receipts: list[ReceiptVerification] = field(
+        default_factory=list[ReceiptVerification]
+    )
     broken_at: int = -1
 
 
--- a/sdk/py/src/agent_receipts/receipt/create.py
+++ b/sdk/py/src/agent_receipts/receipt/create.py
@@ -88,14 +88,13 @@ def create_receipt(input: CreateReceiptInput) -> UnsignedAgentReceipt:
     if input.authorization is not None:
         cs_data["authorization"] = input.authorization
 
-    return UnsignedAgentReceipt(
-        **{
-            "@context": list(CONTEXT),
-            "id": f"urn:receipt:{uuid.uuid4()}",
-            "type": list(CREDENTIAL_TYPE),
-            "version": VERSION,
-            "issuer": input.issuer,
-            "issuanceDate": now,
-            "credentialSubject": CredentialSubject(**cs_data),
-        }
-    )
+    receipt_data: dict[str, Any] = {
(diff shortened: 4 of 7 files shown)


ojongerius:
Responses to review feedback (CI is green):

**1. `github-code-quality` on `taxonomy/config.py` — "`Any` unused":** declining. `Any` is referenced inside `cast("dict[str, Any]", ...)` string forward references (the quoted form is required by ruff's TC006 rule, which fires when `from __future__ import annotations` is in effect). Ruff itself reports `All checks passed!` — the bot doesn't parse type expressions inside quoted `cast()` arguments.

**2. Copilot on `chain.py:32` — `default_factory=list[ReceiptVerification]` "will raise at runtime":** declining. Verified by actually instantiating it:

```
>>> verify_chain([], 'fake-key')
ChainVerification(valid=True, length=0, receipts=[], broken_at=-1)
>>> ChainVerification(valid=True, length=0).receipts
[]
```

Python 3.9+ added `types.GenericAlias.__call__`, so calling `list[ReceiptVerification]()` returns a regular empty list. The dataclass tests and `verify_chain([], ...)` paths exercise this and pass.

**3. Copilot on `signing.py:106` — removing `isinstance(proof_value, str)` is a behavior change:** real point, but declining. The issue (#175) explicitly identified this as `reportUnnecessaryIsInstance` to fix. The field is typed `str` and validated at construction; the only way `proof_value` could be non-str is if a caller bypasses Pydantic via direct mutation, which is a type-contract violation. A verification function silently returning `False` on that masks the bug rather than surfacing it. If we want true defense the right fix is `validate_assignment=True` on the `Proof` model — happy to do that in a follow-up if desired, but it's outside this issue's scope.

**4. Copilot on `.github/workflows/sdk-py.yml` — "needs maintainer review":** noted, informational.

Issue #171, opened 2026-04-14
spec + sdks: document that chain verification does not detect tail truncation


ojongerius:
## Problem

All three SDKs' chain verification (`sdk/go/receipt/chain.go:25-94`; `sdk/ts/src/receipt/chain.ts`; `sdk/py/src/agent_receipts/receipt/chain.py`) checks three things: signature validity, hash linkage, and strict sequence increment. They do NOT check chain completeness.

- **Mid-chain deletion** — caught, because the next receipt's `previous_receipt_hash` no longer matches the new predecessor's hash.
- **Tail truncation** (drop the last N receipts) — **not caught**. `[A, B, C]` derived from the original `[A, B, C, D, E]` verifies as `Valid: true`, `Length: 3`.

This is a protocol-level property, not an SDK bug — the SDK has no way to know how long the chain should be without external information. But:

- Not stated in `spec/spec/agent-receipt-spec-v0.1.md`.
- Not in the `VerifyChain` godoc / docstring / TSDoc.
- No test asserts the property exists or doesn't.
- Callers will reasonably read `Valid: true` as "chain is intact", which is misleading.

## Tasks

- [ ] Add a section to `spec/spec/agent-receipt-spec-v0.1.md` explaining what chain verification does and does not detect, and recommending mitigations (externally-signed chain length, periodic checkpointing, commitment to final hash, etc.).
- [ ] Reflect that language in the `VerifyChain` docstring in all three SDKs.
- [ ] Add a test per SDK that asserts tail truncation verifies `Valid: true` — so future agents don't silently "fix" this by adding a length check without discussion. The tests pin current behaviour and fail loudly if a future maintainer tightens verification.

## Follow-up (separate issue, not this one)

Optional API surface for callers who can supply `ExpectedLength` or `ExpectedFinalHash`. Discuss spec approach first.

---
_Source: internal code-review pass, April 2026._

jagmarques:
Pinning current behavior with a test is the right call. Without it, somebody will read "chain verification" and assume completeness, then quietly add a length check that breaks every existing caller relying on partial verification.

For the spec wording, I'd lean on the "transparency log" framing rather than reinventing it: a self-contained chain can prove ordering and integrity of what's present, but completeness needs an external anchor (signed length, periodic checkpoint, or an inclusion proof against a log). Calling that out explicitly tells implementers what they're getting and what they need to bolt on.

[...]

Issue #153, opened 2026-04-13
Hash server responses in receipts, not just request parameters


ojongerius:
## Problem

Receipts currently store a `parameters_hash` of the tool call arguments (the request), but not the server's response. This means the receipt proves what was *asked* but not what was *returned*.

This creates a gap in the audit trail:

- **Compromised MCP server** — if the server returns modified data (e.g. a `list_issues` response with injected content), the receipt shows the request was legitimate but doesn't capture what came back
- **Proxy tampering** — a modified proxy could alter responses while leaving the request hash intact
- **Debugging** — when investigating an incident, you can see what tool was called with what arguments, but not what the agent actually received and acted on

Without response hashing, the receipt chain proves intent but not outcome.

## Proposal

Add an optional `response_hash` field to the receipt's `outcome` section, containing a SHA-256 hash of the RFC 8785 canonical JSON response. This would make the receipt a complete record of the request-response pair.

### Design considerations

- **Large responses** — some MCP responses are large (e.g. `get_file_contents`). Hashing is cheap but storing full responses would bloat the DB. The hash-only approach keeps receipts small.
- **Redaction ordering** — response hashing must happen *after* secret redaction, so the hash matches what's actually stored/exported
- **Backward compatibility** — `response_hash` should be optional so older receipts remain valid. The schema version should indicate whether response hashing is supported.

## Acceptance criteria

- [ ] Receipt `outcome` section includes a `response_hash` field (SHA-256, RFC 8785 canonical JSON)
- [ ] Response hashing occurs after secret redaction (hash matches the redacted response, not the raw one)
- [ ] `response_hash` is optional — receipts without it are still valid and verifiable
- [ ] `mcp-proxy verify` validates `response_hash` when present, skips when absent
- [ ] Receipt schema version is bumped to reflect the new field
- [ ] Unit tests: response hash generation, verification with and without response hash, redaction-then-hash ordering
- [ ] Integration test: tool call → receipt with response hash → verify chain passes
- [ ] Spec/ADR updated to document the response hashing design decision
- [ ] Docs updated with explanation of what `response_hash` captures and why it matters

jagmarques:
Hashing the response is the right call, but a few details determine whether the field actually buys you the property you want.

[...]

PR #196, opened 2026-04-22
docs(adr): ADR-0008 response hashing and chain completeness

ojongerius:
## Summary

Adds ADR-0008 recording the decision for closing #153 (response hashing) and #171 (chain completeness).

The ADR bundles four sub-decisions intended for a single coordinated `0.2.0` release:

1. **§1 Response hashing (from #153)** — optional `credentialSubject.outcome.response_hash` field, SHA-256 over RFC 8785 canonical JSON, computed after secret redaction. Schema bump `0.1.0` → `0.2.0`.
2. **§2 Documentation of truncation detection limits (Option A, from #171)** — normative spec text + matching SDK docstrings + per-SDK pinning tests asserting `Valid: true` for a truncated chain.
3. **§3 Out-of-band truncation detection (Option B)** — optional `ExpectedLength` / `ExpectedFinalHash` verifier parameters in each SDK. No schema change.
4. **§4 Terminal-marker receipt (Option D)** — optional `credentialSubject.chain.terminal` field (schema-restricted to constant \`true\` or absent, to avoid RFC 8785 canonicalization/signature ambiguity). Adds an automatic 'receipt after terminal' integrity check and an opt-in \`RequireTerminal\` verifier parameter. Schema change, shares the \`0.2.0\` bump with §1.

## Rejected options

An earlier draft considered an in-chain \`length_commitment\` field (Option C). On closer analysis, every concrete interpretation either duplicates \`sequence\`, requires knowledge the issuer does not have, or reduces to a transparency-log protocol — see the 'Why in-chain positive commitment to successors does not work' section. Option C is rejected as security theatre. Transparency-log-style checkpointing remains deferred to a future ADR.

## Test plan

This PR is docs-only. The implementation that follows the ADR should satisfy the acceptance criteria in #153 and the documentation/test tasks in #171.


ojongerius:
Good catch, thanks. Applied the suggestion verbatim in 05d3249 — two affected lines, one fix each: `../../` prefix (file lives in `docs/adr/`) and GitHub's `#L158` anchor format instead of the `:158` editor convention. Verified no other path-style links remain in the file.

ojongerius:
Reopened the Option D decision after a second look (a7e1f81). Original deferral rested on the "most chains will be unmarked so the signal is noisy" argument, which only bites if verifiers *fail* on missing terminal — they don't. Treating `chain.terminal` as a positive claim only (presence = closed, absence = no claim) makes it a cheap, unambiguous in-band signal that pairs with the `0.2.0` bump already motivated by response hashing.

Decision now ships four things in `0.2.0`:

1. §1 `response_hash` (schema)
2. §2 docs + pinning tests for truncation (no schema)
3. §3 `ExpectedLength` / `ExpectedFinalHash` verifier params (no schema)
4. §4 `chain.terminal` boolean + "no receipt after terminal" integrity rule + optional `RequireTerminal` verifier param (schema)

Option C remains rejected; transparency-log checkpointing remains deferred.

ojongerius:
Self-review pass on the Option D amendment caught six issues, pushed as [4748623](https://github.com/agent-receipts/ar/pull/196/commits/4748623):

**Protocol bug (load-bearing)**
- `{"terminal": false}` and `{}` canonicalize differently under RFC 8785 and produce different signatures, despite being semantically identical "no claim" states. Fixed by schema-restricting `chain.terminal` to `"enum": [true]` — explicit `false` is not a valid value; omission is the only way to express "no claim".

**Conceptual / prose (five)**
- Renamed the floor section from "Why in-chain truncation detection does not work" to "Why in-chain *positive* commitment to successors does not work" — terminal is a *negative* commitment, which the issuer always knows at signing time and is therefore outside the floor.
- "Both sub-decisions" → "All four sub-decisions" in the Decision intro.
- Context line was overclaiming "no caller discipline required"; split into the automatic integrity check (no discipline needed) and the opt-in `RequireTerminal` affordance (same discipline level as Option B).
- Tightened what §4 actually detects: integrity check is unconditional; truncation-detection for terminal-ended chains requires caller opt-in.
- Fixed §4.3 imprecision in Consequences — `response_hash` and `chain.terminal` both live under §4.3.2 (`credentialSubject`), not the §4.3 umbrella.

Issue #118, opened 2026-04-12
Use RFC 8785 canonical JSON for parameters_hash


ojongerius:
The proxy and integration tests compute `parameters_hash` via `json.Marshal` → `SHA256Hash`. Since `json.Marshal` on a `map` does not guarantee key order, the hash can be nondeterministic and won't match RFC 8785 (JCS) canonical JSON hashing — which matters for cross-SDK reproducibility.

Replace `json.Marshal` with the receipt SDK's `Canonicalize` function when hashing parameters, so `parameters_hash` is stable and cross-language reproducible.

This affects:
- `mcp-proxy/cmd/mcp-proxy/main.go` (receipt emission)
- Integration tests that mirror the same pattern
- Potentially the TS and Python SDKs if they do the same

Flagged by Copilot review on #113.

Issue #84, opened 2026-04-05
Spec uses validFrom but all SDKs use issuanceDate


ojongerius:
## Problem

The spec (VC Data Model 2.0) uses `validFrom` for the receipt timestamp field, while all three SDKs use `issuanceDate` (VC Data Model 1.x naming):

- **Spec**: `validFrom` throughout — field table (§5), schema, and all examples
- **Go SDK**: `IssuanceDate` with comment: `// issuanceDate follows W3C VC v1 naming; the protocol spec uses this instead of VC v2's validFrom.`
- **Python SDK**: `issuanceDate` in Pydantic models
- **TypeScript SDK**: `issuanceDate` in type definitions
- **Test vectors**: `issuanceDate` in `sdk/py/tests/fixtures/ts_vectors.json`

## Impact

Field names are part of the canonical JSON representation (RFC 8785). A receipt with `validFrom` produces different canonical bytes than one with `issuanceDate`, meaning:

- Any implementation built against the spec will produce signatures/hashes incompatible with all three SDKs
- Cross-language verification will fail if one side follows the spec and another follows the SDK convention

## Resolution needed

Pick one name and align everything:
- **Option A**: Update the spec to use `issuanceDate` (matches SDKs, VC v1 convention)
- **Option B**: Update all three SDKs to use `validFrom` (matches spec, VC v2 convention)

Either way, the schema, test vectors, and all examples need to be updated to match.

Flagged during ADR-0002 review (PR #79).

ojongerius:
**Resolved in v0.2.1 (2026-04-22).**

The spec and schema were updated from `validFrom` → `issuanceDate` to align with all three SDKs. ADR-0009 documents the decision (VC 1.x naming committed; backward-compatible with all shipped receipts since they already emitted `issuanceDate`). No implementation changes were needed.

Tracked in CHANGELOG under `[0.2.1]`.

PR #200, opened 2026-04-22
docs(spec/schema/adr): canonicalisation profile + issuanceDate commitment (ADR-0009)

### LINKED ISSUE #82: Go SDK: sort.Strings does not implement RFC 8785 UTF-16 code unit ordering
## Problem

The Go SDK uses `sort.Strings(keys)` for JSON key ordering during RFC 8785 canonicalization (`sdk/go/receipt/hash.go`). Go's `sort.Strings` sorts lexicographically by byte value (UTF-8 code point order), which is **not** the same as UTF-16 code unit order required by RFC 8785.

For keys containing only BMP characters (U+0000–U+FFFF), UTF-8 byte order and UTF-16 code unit order produce the same result. For supplementary characters (emoji, CJK Extension B, mathematical symbols above U+FFFF), they diverge.

## Impact

- Receipts with keys containing supplementary Unicode characters will canonicalize differently in Go vs Python/TypeScript
- Cross-language signature verification will fail for affected receipts
- Python correctly handles this with `_utf16_sort_key` (encodes to UTF-16-LE for comparison)
- TypeScript gets correct behavior from JS's native UTF-16 string comparison

## Current risk

Low in practice — receipt field names are ASCII. But `credentialSubject` content is user-defined and could contain supplementary Unicode keys.

## Fix

Replace `sort.Strings(keys)` with a custom comparator that sorts by UTF-16 code unit order, similar to the Python SDK's approach. Add test vectors with supplementary Unicode characters to the cross-language test suite.

Flagged during ADR-0002 review (PR #79).

### LINKED ISSUE #83: SDK/schema mismatch: issuanceDate (VC 1.x) vs validFrom (VC 2.0)
## Summary

The JSON schema and spec use `validFrom` (W3C VC Data Model 2.0 naming), but all three SDKs serialize the issuance timestamp as `issuanceDate` (VC 1.x naming).

## Evidence

- **Schema** (`spec/schema/agent-receipt.schema.json`): `validFrom` in `required` array (line 13) and property definition (line 49)
- **Go SDK** (`sdk/go/receipt/types.go:142-144`):
  ```go
  // issuanceDate follows W3C VC v1 naming; the protocol spec uses this
  // instead of VC v2's validFrom.
  IssuanceDate string `json:"issuanceDate"`
  ```
- **TypeScript SDK** (`sdk/ts/src/receipt/types.ts:140`): `issuanceDate: string`
- **Python SDK**: same pattern

## Impact

SDKs produce receipts that fail schema validation on this field. Cross-language verification may also be affected if any implementation switches without the others.

## Suggested resolution

Pick one name and update the other side:
- If `validFrom` (VC 2.0): update all SDKs
- If `issuanceDate` (VC 1.x): update the schema

The ADR-0003 text currently says `validFrom` and notes this mismatch.

Related: ADR-0003 (PR #80)

ojongerius:
## Summary

Spec-and-schema-only change that closes the two ADR-0002 Known Risks that have been open since the beginning, and adds the normative null/optional rule that was missing from the spec entirely. No SDK code changes; those ship in per-SDK follow-up PRs.

- **Commits `issuanceDate` indefinitely** — schema and spec updated to match all three SDKs; `validFrom` (VC 2.0) removed. All existing receipt signatures preserved (every SDK has always emitted `issuanceDate`). Closes #83, #84.
- **Normative null/optional rule** — §7.1.1 added: required-nullable fields (currently only `previous_receipt_hash`) must emit `null`; optional fields must not emit `null` (normalise to absent). Spec-level close of #85; SDK enforcement ships in per-SDK PRs.
- **Schema bumps to 0.2.1** — optional fields that previously declared `["string", "null"]` now declare `"string"` only. Wire format is unchanged for any receipt any SDK has ever produced.
- **ADR-0009** — new decision record covering the two spec-level decisions. Bug fixes (#82, #86, #118, Go HTML-escape) are conformance fixes to ADR-0002, documented in PR descriptions rather than the ADR.
- **Cross-SDK canonicalisation test vectors** (`cross-sdk-tests/canonicalization_vectors.json`) — 39 vectors covering key-sort edge cases (BMP > U+00FF, supplementary plane, surrogate-vs-BMP), number boundaries, string escaping (incl. no HTML escape, no U+2028 escape), null normalisation, `parameters_hash` with non-ASCII keys, full-receipt shapes, and a signature-preservation invariant against the existing v020 vectors.

## What does NOT change

- No SDK code
- No wire format (every shipped receipt verifies identically post-merge)
- No `mcp-proxy` code
- Schema version `0.1.0` and `0.2.0` receipts remain valid

## Test plan

- [x] `cd cross-sdk-tests && go build ./... && go vet ./...` — passes
- [x] No `validFrom` remaining in `spec/` or `cross-sdk-tests/` outside of explanatory notes
- [x] All 8 example JSON files parse and structurally validate manually
- [ ] SDK sessions (Python, TypeScript, Go) will each run `canonicalization_vectors.json` in CI — tracked in #82, #86, #118

## Related

Closes #83, #84. Spec-level close of #85. Follow-up SDK PRs will close #82, #86, #118.
Design note: [docs/adr/0009-canonicalization-and-schema-consistency.md](docs/adr/0009-canonicalization-and-schema-consistency.md)


ojongerius:
Re: [this comment](https://github.com/agent-receipts/ar/pull/200#discussion_r3123053601) about `COMPUTE_AT_COMMIT_TIME` / `SAME_AS_...` placeholders.

Addressed in ea24c49 (now on `cb57779`). Every `COMPUTE_AT_COMMIT_TIME` value is populated with a concrete `sha256:...` computed by Python and cross-verified by Go. The `SAME_AS_receipt_all_optional_absent` markers are kept because they encode the byte-equality *invariant* that those vectors are designed to assert (not a pending value) — `notes_for_implementers` now explicitly documents that.

Also dropped the stale "Fields marked COMPUTE_AT_COMMIT_TIME must be populated…" implementer note.

ojongerius:
Re: [this comment](https://github.com/agent-receipts/ar/pull/200#discussion_r3123053638) — `parameters_hash_*` vectors not actually catching the Go `json.Marshal` bug.

You're right. With keys ÿ (U+00FF) and Ā (U+0100), UTF-8 byte order (C3 BF < C4 80) and UTF-16 code-unit order (0x00FF < 0x0100) both put ÿ first, so the vector didn't distinguish conformant RFC 8785 output from Go's shortcut.

Fixed in ea24c49:

- Renamed the non-ASCII vector to `parameters_hash_supplementary_vs_bmp` and switched the keys to 🎉 (U+1F389) vs U+E000, which *do* diverge:
  - UTF-16 code-unit sort: 🎉's high surrogate 0xD83C < 0xE000 → 🎉 first (RFC 8785 — correct).
  - UTF-8 byte sort: 🎉 = `F0 9F 8E 89` vs U+E000 = `EE 80 80` → U+E000 first (Go default — wrong).
- Canonical bytes pinned to `{"🎉":2,"\ue000":1}`; expectedHash `sha256:772daaff...`.
- Tightened `parameters_hash_ascii` description: it's a parity baseline, not a Go-bug detector (as you noted, ASCII sorts agree).

ojongerius:
Re: [this comment](https://github.com/agent-receipts/ar/pull/200#discussion_r3123053655) — `receipt_hash_vectors` containing schema-invalid receipts (optional fields set to null).

Agreed. After this PR's schema tightening those payloads fail validation, so they can't be treated as conformant receipts. They only exist to exercise the SDK's pre-serialization null→absent sweep.

Clarified in ea24c49:

- `receipt_optional_null_becomes_absent` now carries `"preNormalizationInput": true` and its description leads with **INTENTIONALLY SCHEMA-INVALID caller input**, explicitly stating it's never emitted on the wire.
- Top-level `$comment` warns that some `receipt` payloads carry schema-invalid caller input to test the sweep.
- Added an implementer note: *"Do NOT attempt to schema-validate these `receipt` payloads; canonicalise and hash them and assert the invariant."*

ojongerius:
Re: [this comment](https://github.com/agent-receipts/ar/pull/200#discussion_r3123053680) — ADR-0009 status should be "Accepted".

Changed in cb57779 (the decisions in this ADR are being merged normatively, matching ADR-0008's convention).

ojongerius:
Re: [this comment](https://github.com/agent-receipts/ar/pull/200#discussion_r3123053719) — `schemaVersion: "0.2.1-draft"` should just be `"0.2.1"`.

Changed in ea24c49, and the matching `-draft` line in `notes_for_implementers` dropped.

ojongerius:
Re: [this comment](https://github.com/agent-receipts/ar/pull/200#discussion_r3126948419) — ADR-0009 "all SDKs run it in CI" is aspirational.

Fair point. Reworded in 0d88ec7:

> `cross-sdk-tests/canonicalization_vectors.json` added; each SDK must run it in CI and assert byte-equality on every vector (integration tracked in #82, #86, #118).

The per-SDK CI wiring ships in the follow-up PRs that close those issues.

ojongerius:
Re: [this comment](https://github.com/agent-receipts/ar/pull/200#discussion_r3126948463) — `number_negative_zero` input `-0` may be decoded as integer `0`.

Good catch. Changed `input` to `-0.0` in 0d88ec7, which standard decoders preserve as IEEE 754 negative zero without harness special-casing (Python `json.loads`, JS `JSON.parse`, Go `json.Unmarshal` into `float64` / `interface{}` all round-trip `-0.0` as a float with the negative sign bit). Description rewritten to explain the choice.

Canonical output stays `0`, which is the invariant the vector asserts.

ojongerius:
Re: [this comment](https://github.com/agent-receipts/ar/pull/200#discussion_r3126948487) — spec §7.1.1 says vectors "enforce this invariant in CI" but the CI wiring isn't in this PR.

Same fix as the ADR bullet. Reworded in 0d88ec7:

> Cross-SDK test vectors in `cross-sdk-tests/canonicalization_vectors.json` pin the expected behaviour and are intended to enforce this invariant once integrated into each SDK's CI (tracked in #82, #86, #118).

PR #205, opened 2026-04-22
docs(mcp-proxy): fix installation and integration guides end-to-end

ojongerius:
## Summary

Originally a one-line PATH troubleshooting note; grew to fix every dead end users actually hit while following the install guide end-to-end.

**Installation page**
- Add PATH troubleshooting note for `mcp-proxy` not found after `go install` (bash/zsh + fish, mentions `GOBIN`).
- Replace unrunnable `/path/to/mcp-server.js` placeholder with a runnable `npx @modelcontextprotocol/server-filesystem` smoke test; note the Node.js dependency.

**Integration docs (Claude Code, Claude Desktop, Codex, Overview)**
- Switch from the deprecated `@modelcontextprotocol/server-github` npm package to GitHub's official `github-mcp-server` (`brew install github-mcp-server`) and pass its required `stdio` subcommand.
- Use `$(which mcp-proxy)` / `$(which github-mcp-server)` in CLI examples so paths resolve at paste time regardless of `GOBIN` or install method.
- Add a helper `echo` block above each static JSON/TOML template that prints the exact absolute paths to paste.
- Add an "install the wrapped MCP server" prerequisite to all three integration pages.
- Scope echo helpers to macOS/Linux and add a PowerShell note for Windows readers of the Claude Desktop guide.

## Test plan

- [ ] Render docs locally and confirm the updated code fences render correctly.
- [ ] Walk through the Claude Code install guide top-to-bottom on a clean shell and confirm `claude mcp list` shows the audited server connected.

### DIFF
--- a/site/src/content/docs/mcp-proxy/claude-code.mdx
+++ b/site/src/content/docs/mcp-proxy/claude-code.mdx
@@ -10,13 +10,13 @@ Claude Code is a CLI-based agent that connects to MCP servers for tools like Git
 - [mcp-proxy installed](/mcp-proxy/installation/)
 - A signing key pair generated (see below)
 - Claude Code installed (`npm install -g @anthropic-ai/claude-code`)
-- The MCP server you want to audit. The examples below wrap the GitHub MCP server:
+- The MCP server you want to audit. The examples below wrap GitHub's official MCP server:
 
   ```bash
-  npm install -g @modelcontextprotocol/server-github
+  brew install github-mcp-server
   ```
 
-  This puts an `mcp-server-github` binary on your `$PATH`. Verify with `which mcp-server-github`.
+  This puts a `github-mcp-server` binary on your `$PATH`. Verify with `which github-mcp-server`.
 
 ## Generate a signing key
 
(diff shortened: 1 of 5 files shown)


ojongerius:
Replies to @copilot review:

1. **GOBIN / multi-entry GOPATH** — fair catch on accuracy. Added a short factual mention: _"writes binaries to `$GOBIN` when it's set, otherwise to `$(go env GOPATH)/bin`"_. Skipped the suggested heredoc/IFS snippet — it's too elaborate for a troubleshooting note targeted at users who just need to get one binary on PATH, and users with non-default `GOBIN` or multi-entry `GOPATH` already know how to read `go env` themselves.

2. **Mixed bash/fish inside a `bash` fence** — fixed. Split into two fences with `bash` and `fish` tags respectively.

3. **"Unknown command" is fish-specific** — fixed. Reworded to "_bash/zsh say `command not found`; fish says `Unknown command`_" so both shells' users recognise their error.

Pushed in a99d128.

ojongerius:
Second+third Copilot round replies:

1. **`$(go env GOPATH)/bin/mcp-proxy` ignores GOBIN** (flagged in `claude-code.mdx`, `claude-desktop.mdx`, `codex.mdx` — both heredoc/CLI and helper `echo` blocks) — fixed. Swapped all 5 instances to `$(which mcp-proxy)` (and `$(which github-mcp-server)` where applicable), which naturally handles `GOBIN`, multi-entry `GOPATH`, and users who installed via any other mechanism. The installation page's troubleshooting note still uses `$(go env GOPATH)/bin` since that section is precisely about the case where `mcp-proxy` isn't on `$PATH` yet (so `which` would be empty).

2. **claude-desktop.mdx mentions Windows but the echo snippet is bash-only** — fixed. Scoped the echo block to macOS/Linux and added a PowerShell one-liner (`(Get-Command mcp-proxy).Source`, `$env:USERPROFILE`, `(Get-Command github-mcp-server).Source`) for Windows users.

3. **npx smoke test requires Node** — fixed. Added an inline note: _"requires Node.js/`npx` — if you don't have it, substitute any MCP server command you already have"_.

4. **PR scope drifted from title** — fixed. Updated PR title and description to cover the full end-to-end doc fixes (install PATH note, npx smoke test, github-mcp-server swap, `stdio` subcommand, `$(which …)` everywhere, wrapped-server install prereq, Windows note).

Pushed in be0daeb.

PR #209, opened 2026-04-23
fix(mcp-proxy/classifier): add suffix matching for resource-first tool names

### LINKED ISSUE #208: mcp-proxy: classifier misses resource-first tool names (e.g. github-mcp-server)
## Problem

`audit.ClassifyOperation` in `mcp-proxy/internal/audit/classifier.go` only matches **action-first** tool names (`get_*`, `list_*`, `create_*`, `delete_*`, …).

GitHub's official [`github-mcp-server`](https://github.com/github/github-mcp-server) — which we now recommend in our integration docs after the previous server was deprecated — uses **resource-first** naming: `pull_request_read`, `pull_request_create`, `repository_delete`, etc.

Result: every call through the proxy with no taxonomy entry falls through to `"unknown"` and defaults to medium risk, regardless of the real operation.

### Example receipt

```json
{
  "action": {
    "type": "unknown",
    "tool_name": "pull_request_read",
    "risk_level": "medium",
    "target": { "system": "github" }
  }
}
```

`pull_request_read` should classify as `read` / low risk.

## Proposed fix

Extend `ClassifyOperation` to also check **suffixes** after the existing prefix checks:

- `_read` → read
- `_create` / `_update` / `_write` → write
- `_delete` / `_remove` → delete
- `_run` / `_exec` → execute

Prefix checks should continue to run first so nothing already-working regresses. Taxonomy mappings remain the escape hatch for per-tool overrides (e.g. a `_read` tool that should actually be high-risk).

### Why suffix matching over shipping a taxonomy file

Considered shipping a taxonomy JSON for github-mcp-server instead. Decided against as the default:

- A taxonomy file needs a new entry for every new tool the server adds — silent regression to `unknown` otherwise.
- `resource_action` is a common naming pattern, not github-mcp-server-specific. Fixing the classifier helps the long tail of servers that follow it.
- Taxonomy is still available for users who want precise per-tool control.

## Scope

- Update `mcp-proxy/internal/audit/classifier.go`
- Extend `mcp-proxy/internal/audit/classifier_test.go` with cases for `pull_request_read`, `pull_request_create`, `repository_delete`, and an ambiguous case like `repository_read_only_mode` to lock in behavior.
- Confirm risk-level alignment in the fallback branch around `mcp-proxy/cmd/mcp-proxy/main.go:493` still works (it keys off the returned `actionType`, so should be automatic).

ojongerius:
## Summary

- Extends `ClassifyOperation` in `classifier.go` to check `_read`/`_create`/`_update`/`_write`/`_delete`/`_remove`/`_run`/`_exec` **suffixes** after the existing prefix checks
- Prefix checks run first — no regression to already-working tool names
- Taxonomy mappings remain the per-tool override escape hatch

## Test plan

- [ ] `pull_request_read` → `read`
- [ ] `pull_request_create` → `write`
- [ ] `repository_delete` → `delete`
- [ ] `repository_read_only_mode` → `unknown` (ambiguous; does not end with `_read`)
- [ ] Existing prefix-based cases unchanged (`read_file`, `delete_file`, etc.)
- [ ] `go test ./internal/audit/...` passes

Closes #208

### DIFF
--- a/mcp-proxy/internal/audit/classifier.go
+++ b/mcp-proxy/internal/audit/classifier.go
@@ -38,6 +38,36 @@ func ClassifyOperation(toolName string) string {
 		}
 	}
 
+	// Suffix checks for resource-first naming (e.g. github-mcp-server's pull_request_read).
+	// Same priority order as prefix checks: delete > execute > write > read.
+	deleteSuffixes := []string{"_delete", "_remove"}
+	for _, s := range deleteSuffixes {
+		if strings.HasSuffix(lower, s) {
+			return "delete"
+		}
+	}
+
+	execSuffixes := []string{"_run", "_exec"}
+	for _, s := range execSuffixes {
+		if strings.HasSuffix(lower, s) {
+			return "execute"
+		}
+	}
+
+	writeSuffixes := []string{"_create", "_update", "_write"}
+	for _, s := range writeSuffixes {
+		if strings.HasSuffix(lower, s) {
+			return "write"
+		}
+	}
+
+	readSuffixes := []string{"_read"}
+	for _, s := range readSuffixes {
+		if strings.HasSuffix(lower, s) {
+			return "read"
+		}
+	}
+
 	return "unknown"
 }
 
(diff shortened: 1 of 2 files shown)


ojongerius:
Closing in favour of #210 which has the same fix with a more complete test suite.

ojongerius:
Closing in favour of #210 which has the same fix with a more complete test suite.

PR #211, opened 2026-04-23
feat(mcp-proxy): default -db/-receipt-db to $HOME/.agent-receipts

### LINKED ISSUE #207: mcp-proxy: default -db / -receipt-db to absolute paths
`mcp-proxy` defaults `-db` and `-receipt-db` to relative paths (`audit.db`, `receipts.db`). When MCP clients (Claude Desktop, Code, Codex) spawn the proxy, cwd is unwritable and the proxy crashes with `unable to open database file (14)` before responding to `initialize`.

Docs currently work around this by requiring absolute paths in every example (#205, #206). Better: fix the defaults.

**Fix:**

1. In `mcp-proxy/cmd/mcp-proxy/main.go` (~line 78), default `-db` to `$HOME/.agent-receipts/audit.db` and `-receipt-db` to `$HOME/.agent-receipts/receipts.db`.
2. `os.MkdirAll` the parent directory (0o700) before opening SQLite.
3. Update flag help text and `configuration.mdx`.
4. Drop the flags from minimal integration examples once shipped.

**Acceptance:**

- [ ] `mcp-proxy <cmd>` succeeds from any cwd with no db flags.
- [ ] `~/.agent-receipts/` created on first run.
- [ ] Docs updated.

ojongerius:
## Summary

- Default `-db` and `-receipt-db` to `$HOME/.agent-receipts/{audit,receipts}.db`, creating the parent directory with mode 0700 on first open.
- Apply the same defaults to `list`/`inspect`/`verify`/`export`/`stats`/`timing` so the audit CLI works from any cwd.
- Drop the now-redundant `-db`/`-receipt-db` flags from minimal integration examples and update the configuration + CLI references.

## Motivation

MCP clients (Claude Desktop, Claude Code, Codex) spawn the proxy with an unwritable cwd, so the previous relative defaults caused `unable to open database file (14)` before `initialize`. Every integration guide worked around it by pinning absolute paths — fix the default instead.

Closes #207.

## Test plan

- [x] `go vet ./...`, `go build ./...`, `go test ./...`
- [x] `go test -tags=integration .`
- [x] `go test -tags=e2e .`
- [x] Manually wire an MCP client with no `-db`/`-receipt-db` flags and confirm `~/.agent-receipts/` is created on first use

### DIFF
--- a/mcp-proxy/cmd/mcp-proxy/cli.go
+++ b/mcp-proxy/cmd/mcp-proxy/cli.go
@@ -23,7 +23,7 @@ func openReceiptStore(path string) *store.Store {
 
 func cmdList(args []string) {
 	fs := flag.NewFlagSet("list", flag.ExitOnError)
-	db := fs.String("receipt-db", "receipts.db", "Receipt store path")
+	db := fs.String("receipt-db", defaultDBPath("receipts.db"), "Receipt store path")
 	chainID := fs.String("chain", "", "Filter by chain ID")
 	riskLevel := fs.String("risk", "", "Filter by risk level")
 	actionType := fs.String("action", "", "Filter by action type")
@@ -83,7 +83,7 @@ func cmdList(args []string) {
 
 func cmdInspect(args []string) {
 	fs := flag.NewFlagSet("inspect", flag.ExitOnError)
-	db := fs.String("receipt-db", "receipts.db", "Receipt store path")
+	db := fs.String("receipt-db", defaultDBPath("receipts.db"), "Receipt store path")
 	pubKeyPath := fs.String("key", "", "Public key (PEM file) for signature verification")
 	fs.Parse(args)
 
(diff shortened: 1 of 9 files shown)


ojongerius:
Closing the loop on the two outstanding Copilot suggestions on `4d3b566` ([thread 1](https://github.com/agent-receipts/ar/pull/211#discussion_r3128089010), [thread 2](https://github.com/agent-receipts/ar/pull/211#discussion_r3128089039)) — both are valid but I deferred them as out of scope for this PR:

**1. `ensureDBDir` doesn't tighten perms on a pre-existing directory.** True — `os.MkdirAll` is a no-op when the directory already exists, so a `~/.agent-receipts/` someone created manually at `0o755` stays `0o755`. In practice the proxy creates the directory itself on first run (with `0o700`), so this only matters when a user has already set up the directory with broader perms — at which point silently overwriting their choice would be more surprising than respecting it. If we want to harden this, the right shape is a startup warning when existing perms are broader than `0o700`, not a silent `Chmod`. Tracking as a follow-up rather than expanding this PR.

**2. `defaultDBPath` fallback is silent.** Also true — if `os.UserHomeDir()` fails entirely (e.g. a minimal container with no `/etc/passwd` entry), we fall back to a bare `audit.db`/`receipts.db` and the proxy can still hit the original "unable to open database file" error this PR set out to fix. The realistic fix is a fail-loud check after flag parsing in `serve()` and each CLI subcommand. Worth doing, but it's an additional behavior change on top of an already-iterated PR; better as its own change so the diff is reviewable on its own merits.

Merging now; will open issues for both.

Issue #172, opened 2026-04-14
mcp-proxy: add signal handling and graceful shutdown


ojongerius:
## Problem

`mcp-proxy/cmd/mcp-proxy/main.go` has no `signal.Notify` wiring. On SIGTERM/SIGINT:

- Deferred `Close()` on audit DB and receipt store fires on normal return from `main` — OK.
- The HTTP approval server goroutine (`startHTTPServer`) blocks on `http.Serve` with no `Shutdown` — goroutine leak until process exit; no grace for in-flight requests.
- Proxy STDIO pump goroutines may hold buffered writes on interrupt.
- In-flight approval workflows have no deterministic cancellation path.

For a proxy sitting in front of every agent action, "crash-safe" (rely on SQLite WAL) is probably fine in practice, but "graceful" is better and gives observability (a structured shutdown log line instead of abrupt disappearance).

## Investigate first

Don't jump to a fix — the right shape depends on what's actually broken vs. theoretically untidy:

- Does SIGTERM actually leave the receipt chain in a bad state? (Probably not — SQLite WAL protects write atomicity.)
- Do HTTP-server goroutines leak in practice, or does process exit reap them cleanly?
- What happens to a pending approval if the user's UI is mid-POST when the proxy dies?

Ground the fix in measured behaviour.

## Likely shape

- Root `context.Context` threaded through audit store, HTTP server, proxy loops.
- `signal.NotifyContext(ctx, SIGTERM, SIGINT)` in `main`.
- `http.Server` (not bare `http.Serve`) with `Shutdown(ctx)` and short grace window.
- Pending approval channels closed deterministically.

## Non-goals

- Draining (waiting for every in-flight tool call to finish).
- Systemd-style readiness signaling.

## Acceptance

- `kill -TERM $pid` during a synthetic workload leaves the audit DB consistent (covered by test).
- HTTP server goroutine exits within a bounded grace period.
- Logs distinguish graceful vs. forced shutdown.

---
_Source: internal code-review pass, April 2026._

Issue #157, opened 2026-04-13
Add proxy-side request logging for debugging denied/failed calls


ojongerius:
## Problem

When a tool call is denied upstream (by the MCP client) before reaching the proxy, there's no record of it anywhere. The proxy's audit DB only captures calls that made it through, and receipts are only generated for completed calls.

During a real session, 3 out of 10 `create_issue` calls were denied with MCP error `-32002` (client-side denial). The proxy's `tool_calls` table had no record of these — they simply vanished. The only way to know they happened was the error message returned to the user.

Even for calls that *do* reach the proxy, there's no stderr or file-based log for operational debugging — only the structured audit DB and receipt chain.

### Why this matters

- **Debugging** — when a tool call fails silently, there's no way to diagnose why without adding logging yourself
- **Completeness** — the audit trail has invisible gaps for denied calls
- **Operations** — in production, you need logs for alerting, dashboards, and incident investigation

## Proposal

Add a tiered logging mechanism to the proxy:

**Default level** (low sensitivity) — log to stderr or a configurable log file:
- Tool name
- Timestamp
- Outcome (pass / deny / error / timeout)
- Duration
- MCP request ID (for correlation)

Example: `2026-04-13T07:48:23Z | create_issue | denied by client | no receipt generated`

**Verbose/debug level** (high sensitivity) — additionally log:
- Full tool call arguments (with the same redaction patterns used for receipts)
- Error details and MCP error codes
- MCP protocol-level metadata (JSON-RPC message IDs)

This keeps the default log safe to ship to a SIEM or share in a bug report, while verbose mode is available for local debugging.

### Sensitivity considerations

- Default log level contains no sensitive data (tool names and timestamps only)
- Verbose level must apply the same redaction patterns as receipt storage (#150)
- Log file location should default to `~/.agent-receipts/proxy.log` with configurable path
- Log rotation should be considered (or documented as the user's responsibility)

## Acceptance criteria

- [ ] Proxy emits structured log lines to stderr by default (one line per tool call)
- [ ] Default log level includes: timestamp, tool name, outcome, duration, request ID
- [ ] `-log-level verbose` flag adds redacted arguments and error details
- [ ] `-log-file` flag redirects log output to a file instead of stderr
- [ ] Verbose mode applies the same redaction patterns as receipt storage
- [ ] Log format is parseable (structured JSON or key=value pairs, not free-form text)
- [ ] Unit tests: log output for successful call, failed call, timeout, each log level
[...]

Issue #149, opened 2026-04-13
Document proxy crash/timeout behaviour


ojongerius:
## Problem

When the proxy crashes or hangs mid-call, it's unclear what happens to the MCP client (e.g. Claude Desktop). Several failure scenarios are undocumented:

1. **Proxy process dies mid-request** — does the client get an error, or does it hang waiting for a response on stdin?
2. **Proxy hangs (e.g. deadlock in receipt signing)** — is there a timeout? Does the client retry?
3. **Partial receipt** — if the proxy crashes after forwarding the request to the MCP server but before writing the receipt, is the action recorded or lost?
4. **Receipt DB locked** — if SQLite is locked (e.g. another process reading), does the proxy block the tool call or fail?

## Proposal

- Investigate and document the actual behaviour for each scenario above
- Add a configurable timeout for upstream MCP server responses (default: 30s?)
- Ensure partial/orphaned receipts are handled (either completed on restart or marked as incomplete)
- Add a troubleshooting section to the docs covering these failure modes

## Acceptance criteria

- [ ] Each failure scenario (process death, hang, partial receipt, DB lock) is tested and behaviour is documented
- [ ] Proxy has a configurable timeout flag (`-timeout`) for upstream MCP server responses
- [ ] When the upstream server times out, the proxy returns a well-formed MCP error to the client (not a hang)
- [ ] Partial receipts are either: (a) not written until the full round-trip completes, or (b) marked with `status: incomplete` and cleaned up
- [ ] SQLite DB locking does not block tool calls — proxy either retries with backoff or queues the receipt write
- [ ] Unit tests cover each failure mode (mock upstream crash, timeout, DB lock)
- [ ] Troubleshooting section added to docs with symptoms and remediation for each scenario

PR #221, opened 2026-04-23
feat(mcp-proxy): surface policy + approver state to prevent silent rejections

ojongerius:
## Summary

Four coordinated observability improvements so `-32002` / `-32003` debugging doesn't take hours. Context in [#157 comment](https://github.com/agent-receipts/ar/issues/157#issuecomment-4308580739).

1. **Startup banner** — `[INFO]`/`[WARN]` one-liner on stderr summarising loaded rules and approver URL. WARN fires when pause rules exist but no approver is wired up (the #1 silent-failure mode). JSON companion line for tooling.
2. **Distinguished error paths** — new `-32003` code + `no approver configured ...` message for the no-approver case; timeout/deny keep `-32002` with their wording. `-http=none` explicitly opts out. One structured `policy_event tool=... rule=... outcome=... duration_ms=...` log line per pause/block event.
3. **Persist blocked/rejected rows** — `tool_calls` now records calls that never reached upstream, so `mcp-proxy timing` reflects this session's activity instead of only historical pass-through calls.
4. **`mcp-proxy doctor` subcommand** — dry-run diagnostic (loads rules, probes optional `--approver` URL, lists issues, non-zero exit + JSON output). Pure `DiagnoseConfig` function is test-friendly and usable from install scripts / CI.

Also adds a per-tool `pass / flag / pause / block / rejected` breakdown table to `mcp-proxy timing`.

Refs #157 (closes most of it — verbose/`-log-file` left for a follow-up), #149 (failure modes now distinguishable), #172 (structured events give shutdown something to drain).

## Test plan

- [x] `go test ./...` — passes (new tests for Describe, banner INFO/WARN, distinguished errors, rejected-row persistence, doctor healthy / no-approver / unreachable paths).
- [x] `go vet ./...` — clean.
- [x] `gofmt -l` — clean.
- [x] `mcp-proxy doctor` manual smoke test (default rules, unreachable URL).
- [ ] Follow-up in a separate PR: verbose log level + `-log-file` (remainder of #157).

### DIFF
--- a/mcp-proxy/cmd/mcp-proxy/cli.go
+++ b/mcp-proxy/cmd/mcp-proxy/cli.go
@@ -4,10 +4,13 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/agent-receipts/ar/mcp-proxy/internal/audit"
+	"github.com/agent-receipts/ar/mcp-proxy/internal/policy"
 	"github.com/agent-receipts/ar/sdk/go/receipt"
 	"github.com/agent-receipts/ar/sdk/go/store"
 )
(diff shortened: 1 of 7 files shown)


ojongerius:
Addressed all nine Copilot review comments in ff62098:

1. **`emitStartupBanner` takes `approverDisabled`** — `-http=none` opt-out shows `INFO approver: disabled` instead of the WARN path reserved for the misconfig case.
2. **`emitPolicyEvent` quotes strings with `%q`** — rule/tool names with spaces or `=` parse unambiguously.
3. **`TimingStats.Total` counts every row; `TimedTotal` is the duration subset** — blocked/rejected calls now show up in the total; percentiles/averages still use timed rows.
4. **`captureStderr` uses `defer w.Close()`** — reader goroutine unblocks on panic/Fatal.
5. **Banner splits pause and block into separate `(…)` groups** — operator can see which rules each count refers to.
6. **`DiagnoseConfig` doc reworded** — no longer claims "pure"; it mutates nothing but reads config + probes the approver.
7. **`probeApprover` tries HEAD first, falls back to GET on 405/501** — lighter probe when the approver supports it.
8. **`policyActionBreakdown` uses `sort.Slice`** — replaces the hand-rolled insertion sort.
9. **`TestDescribe` derives expected counts from `len(DefaultRules())`** — default ruleset changes won't break the test.

Also added `TestStartupBannerNoWarnWhenApproverExplicitlyDisabled` to cover the new opt-out path. `go test ./mcp-proxy/...`, `go vet`, and `gofmt` all clean.

ojongerius:
Addressed the six new Copilot comments from the ff62098 review in 0f02e33:

1. **`recordRejectedToolCall` reuses `requestedAt`** — the block/pause paths now pass the same timestamp captured when the pendingCall entry was created, so rejected rows align in time with the original request instead of drifting to rejection time.
2. **Only `-http=none` counts as the explicit opt-out** — empty `-http` now falls back to "not configured" so the banner still WARNs when pause rules exist. Catches the mis-parsed flag case Copilot flagged.
3. **Banner wording** — "`%d rules enabled`" (was "rules loaded") plus a "`(N disabled)`" suffix when `TotalRules != EnabledRules`.
4. **`captureStderr` closes the write end once** — `fn()` runs inside an IIFE with a deferred `w.Close()`, removing the double close that followed the deferred one.
5. **`policyActionBreakdown` ties break on ToolName** — `sort.SliceStable` with a secondary key so CLI output and JSON snapshots are deterministic.

`go test ./mcp-proxy/...`, `go vet`, `gofmt` all clean.

PR #222, opened 2026-04-23
fix(mcp-proxy): SQL-mutation risk heuristic no longer false-positives on prose

ojongerius:
## Bug

The `ScoreRisk` function in `mcp-proxy/internal/audit/classifier.go` serialised the entire tool-call arguments map to JSON and searched the raw string for SQL mutation keywords (`drop`, `delete`, `update`, `truncate`, `alter`). Any tool whose arguments contained those words in natural-English prose tripped the heuristic.

**Concrete example from production:**

```
create_pull_request | write | 50 | rejected |
  ["write operation", "SQL mutation without WHERE clause"]
```

A `create_pull_request` call with `title: "docs: update ecosystem framing"` scored:
- +20 write operation
- +30 SQL mutation (the word "update " appears in the title, no "where" nearby)
- **Total: 50 — hits the default `pause_high_risk` threshold**

Sibling tools like `create_branch` / `push_files` escaped only because their argument values happened not to contain the exact tokens.

## Fix

Gate the mutation check on SQL-context argument keys (`sql`, `query`, `statement`, `command`, case-insensitive) rather than scanning the full JSON blob. A new helper `sqlValContainsMutation` walks the arguments map recursively so nested argument objects are also covered. The WHERE-clause guard and the +30 magnitude are unchanged.

Key properties of the fix:
- `ScoreRisk` signature is unchanged — fully backward compatible
- Recursive walk handles nested args (e.g. `{"params": {"statement": "…"}}`)
- Values under non-SQL keys (title, body, branch, message, …) are never inspected
- Case-insensitive key and value matching is preserved

## Tests added (`TestScoreRisk_SQLMutationHeuristic`)

| # | Input | Expected score | SQL mutation reason? |
|---|-------|----------------|----------------------|
| 1 | `create_pull_request` title `"docs: update ecosystem framing"` | 20 (write only) | absent |
| 2 | `{"query": "UPDATE users SET x=1"}` | 60 (execute + SQL) | present |
| 3 | `{"query": "UPDATE users SET x=1 WHERE id=1"}` | 30 (execute only) | absent |
| 4 | `{"sql": "DROP TABLE foo"}` | 60 (execute + SQL) | present |
| 5 | `{"params": {"statement": "DELETE FROM t"}}` (nested) | 60 (execute + SQL) | present |
| 6 | `{"title": "docs: delete stale section"}` | 20 (write only) | absent |

The existing `TestScoreRisk` case at line 69 (`{"query": "drop table secrets"}`) continues to pass.

### DIFF
--- a/mcp-proxy/internal/audit/classifier.go
+++ b/mcp-proxy/internal/audit/classifier.go
@@ -1,8 +1,15 @@
 package audit
 
 import (
+	"regexp"
 	"strings"
 )
+
+// whereClauseRe matches the SQL keyword "where" with token boundaries, so
+// words like "somewhere" or "whereabouts" do not falsely suppress the
+// mutation signal. Case-insensitive; boundaries are start/end of string or
+// any non-letter character.
+var whereClauseRe = regexp.MustCompile(`(?i)(^|[^a-z])where([^a-z]|$)`)
 
 // ClassifyOperation determines the operation type from a tool name.
 func ClassifyOperation(toolName string) string {
@@ -79,7 +86,9 @@ func sqlValContainsMutation(args map[string]any, sqlContextKeys, mutations []str
 			for _, ctxKey := range sqlContextKeys {
 				if kLower == ctxKey {
 					s := strings.ToLower(val)
-					if strings.Contains(s, "where") {
+					// Token-boundary check: "somewhere" / "whereabouts"
+					// must not count as a WHERE clause.
+					if whereClauseRe.MatchString(s) {
 						break
 					}
 					for _, m := range mutations {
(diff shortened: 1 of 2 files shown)


ojongerius:
Addressed both Copilot review comments in a770569:

1. **Array recursion** (classifier.go:100): `sqlValContainsMutation` now recurses into `[]any` as well as `map[string]any`, via a new `sqlSliceContainsMutation` helper. SQL-context keys nested inside arrays (e.g. `{"batch": [{"query": "UPDATE ..."}]}`) are now detected — fixes the behavioral regression vs the prior full-JSON scan.

2. **Word-boundary WHERE** (classifier.go:84): replaced `strings.Contains(s, "where")` with a regex `(?i)(^|[^a-z])where([^a-z]|$)`. Strings like `"find somewhere else"` no longer suppress a real mutation signal. Chose the regex form Copilot suggested over space-padding because it cleanly handles end-of-string and punctuation boundaries.

New test coverage:
- `{"batch": [{"query": "UPDATE ..."}]}` — triggers +30 (array recursion).
- `{"params": {"items": [{"sql": "DROP TABLE foo"}]}}` — triggers +30 (nested map-in-array).
- `{"query": "UPDATE ... SET note='find somewhere else'"}` — triggers +30 (somewhere no longer masks).
- `{"query": "UPDATE ... WHERE id=1"}` and `{"query": "UPDATE ... where id=1"}` — still NOT triggered.

`go vet ./...` and `go test ./...` pass in mcp-proxy.

PR #251, opened 2026-04-26
chore(sdk/ts): untrack auto-generated version.ts, add to .gitignore

### LINKED ISSUE #250: chore(sdk/ts): version.ts is stale (0.3.0) — out of sync with package.json (0.4.0)
## Problem

`sdk/ts/src/version.ts` is currently committed at `VERSION = "0.3.0"` but `sdk/ts/package.json` is at `0.4.0`. The file header says it's auto-generated by `scripts/sync-version.mjs` and "cannot drift from the published version", but it has — the 0.4.0 release commit ([350bb13](https://github.com/agent-receipts/ar/commit/350bb13)) didn't run sync-version.

`pnpm test` and `pnpm typecheck` both regenerate it via the `pretest` / `pretypecheck` hook, so any branch that runs CI locally produces an unrelated diff in this file.

## Fix

Either:
1. Commit the regenerated `version.ts` (one-line change: `0.3.0` → `0.4.0`), or
2. Untrack `version.ts` and add it to `.gitignore` since it's auto-generated from `package.json`.

Option 2 is cleaner — the file is by definition derivable from `package.json`, so committing it just creates drift opportunities like this one.

## Context

Spotted while preparing #249 — the `pretest` hook produced an unrelated diff that I had to revert before committing. Worth fixing so future contributors don't hit the same papercut.

ojongerius:
## Summary

- Adds `src/version.ts` to `sdk/ts/.gitignore` so the auto-generated file is never committed
- Removes `sdk/ts/src/version.ts` from git tracking (`git rm --cached`)
- The file continues to be regenerated before every `build`, `test`, and `typecheck` via the existing `sync-version` script

## Why

`version.ts` is auto-generated from `package.json` by `scripts/sync-version.mjs`. Committing it creates drift opportunities — the 0.3.0/0.4.0 mismatch in #250 is a direct example. Since the file is fully derivable from `package.json`, tracking it in git adds risk with no benefit.

Closes #250


ojongerius:
Good catch, @copilot-pull-request-reviewer — switched to `prepare` in 11f06c5. Verified locally: deleting `src/version.ts` and running `pnpm install` regenerates it via the `prepare` script, and `prepare` won't run for downstream consumers installing the published tarball.

Issue #263, opened 2026-04-28
mcp-proxy: cache OAuth tokens for remote MCP servers (blocked on #227)


ojongerius:
## Background

Spun out of #262 (originally framed as an OAuth token caching problem; the actual issue there turned out to be an `-http` port collision in mcp-proxy's approval listener, unrelated to OAuth).

This issue tracks the **separate, still-valid** idea of OAuth token caching for remote MCP servers.

## Why it's worth doing

When `mcp-proxy` wraps a remote OAuth-protected MCP server (e.g. Atlassian, Slack, Linear), the OAuth flow today is owned entirely by the spawned `mcp-remote` child process. That has two visible costs:

1. **Cold-start browser round-trip on every fresh session** — slow and disruptive UX.
2. **Per-session OAuth state** — `mcp-remote` runs its own callback listener on a port; concurrent sessions can collide there too (separate from #262's issue, which is mcp-proxy's own approval port).

Caching the token on disk after first auth, keyed by `-name`, would let subsequent sessions skip the browser dance entirely:

1. **Valid cached token** → inject directly, no callback listener, no browser
2. **Expired token + refresh token** → silent refresh, update cache
3. **No token / refresh failed** → full OAuth dance, write cache on success

## The architectural question

Token caching only makes sense in mcp-proxy if mcp-proxy owns the OAuth flow. Today it doesn't — `mcp-remote` does. So this issue is **blocked on / scoped against** #227 (native HTTP/SSE transport for remote MCP servers).

Two paths forward:

- **(a) Defer until #227.** Once mcp-proxy talks HTTP/SSE directly to remote servers, it holds the bearer token and token caching becomes its problem. Token-cache design lives in the #227 ADR.
- **(b) Investigate `mcp-remote`'s existing cache.** `mcp-remote` already caches under `~/.mcp-auth/`. Worth checking whether the cache key correctly isolates per-server and survives concurrent sessions. If it works, this is a docs/known-good-pattern fix rather than new code.

## Proposed scope

Token cache layout (when we own it):

- `~/.agent-receipts/oauth/<name>.token.json` — keyed by `-name`
- Permissions `0600`, sits alongside the existing PEM key
- Contents: `{access_token, refresh_token, expires_at, issuer, scopes}`

Decision tree on startup:

1. Read cache → if valid, inject token, skip OAuth listener
2. If expired but refresh token present → silent refresh via stored token endpoint, update cache
[...]

Issue #227, opened 2026-04-24
mcp-proxy: native HTTP/SSE transport for remote MCP servers


ojongerius:
Follow-up to #226 (Option A). #226 itself is being addressed via Options B (docs for the `mcp-remote` composition pattern) and C (bundled taxonomies for prominent remote servers like Atlassian). This issue tracks the larger architectural change.

## Goal

Allow `mcp-proxy` to talk to a remote MCP server directly, without composing `mcp-remote` in front of it:

```sh
mcp-proxy -name atlassian https://mcp.atlassian.com/v1/mcp
```

…with auth header passthrough:

```sh
mcp-proxy -name atlassian \
  -header "Authorization:Bearer $TOKEN" \
  https://mcp.atlassian.com/v1/mcp
```

## Why this is its own issue

The current proxy is a clean stdio bridge ([mcp-proxy/internal/proxy/proxy.go](https://github.com/agent-receipts/ar/blob/main/mcp-proxy/internal/proxy/proxy.go)) — `exec.Command` plus bidirectional pipe. Adding a remote transport is a meaningful architectural shift, not a small flag addition. It needs:

- A second transport implementation (Streamable HTTP and/or SSE)
- A decision about OAuth ownership (does the proxy run the OAuth flow itself, or only forward static credentials and delegate the flow to a sidecar?)
- Reconnect / resumption semantics for SSE
- Header-flag spec, including how multiple `-header` values compose
- Redaction policy for bearer tokens flowing through the proxy (the upside of native transport is that the proxy *sees* the token and can attribute receipts to a token identity; the cost is one more high-value secret to redact)

## Suggested next step

Draft an ADR before any implementation. Key open questions for the ADR:

1. **OAuth ownership**: full client (proxy runs the auth dance) vs. credential-passthrough-only (caller supplies the token, proxy forwards). The latter is far smaller and arguably better separation of concerns.
2. **Transport scope**: Streamable HTTP only (newer MCP spec), SSE only (what most current remote servers expose), or both.
3. **Discovery**: do we keep the positional-arg shape (`mcp-proxy ... <command-or-url>`) and detect URLs, or introduce a dedicated subcommand?
4. **Receipt identity**: when the proxy holds the bearer token, can/should the operator/principal fields on the receipt reflect token identity?

## Out of scope here

- Documentation of the existing `mcp-remote` composition pattern — covered in #226 / Option B.
- Bundled taxonomies for Atlassian/Slack/etc. — covered in #226 / Option C.

PR #266, opened 2026-04-28
fix(mcp-proxy): make approval listener opt-in (closes #262)

### LINKED ISSUE #262: mcp-proxy: -http port collision crashes second concurrent session
## Problem

When multiple Claude Code sessions are open simultaneously, each spawns its own `mcp-proxy` subprocess. If the user's config hardcodes `-http 127.0.0.1:8082` (as our docs currently suggest), every instance tries to bind the same port. Only one wins; the rest crash on startup.

**Observed symptom:** `/mcp` shows `✘ failed` for `atlassian-audited` in all but one session. Killing all `mcp-proxy` processes and reconnecting only helps temporarily — whichever session reconnects second will fail again.

## Root cause (corrected)

The `-http` flag is **not** an OAuth callback listener. It's the **policy-approval endpoint** (`POST /api/tool-calls/{id}/approve|deny` for `pause`-action rules). See `mcp-proxy/cmd/mcp-proxy/main.go:97`.

The collision sequence:

1. With no `-rules` flag, `main.go:198` loads `policy.DefaultRules()`.
2. Defaults include `pause_high_risk` (`internal/policy/defaults.go:18`), so `engine.HasPauseRules() == true`.
3. `main.go:252` calls `net.Listen("tcp", *httpAddr)`.
4. On session 2, the bind fails → `log.Fatalf` at `main.go:255` → `mcp-proxy` exits before the stdio bridge starts → Claude shows `✘ failed`.

The default for `-http` is already `127.0.0.1:0` (random port). The collision only happens because `8082` is hardcoded in user configs (sourced from our docs).

## Fix

Two complementary changes:

### 1. Doc fix (primary)

Update every docs example that uses `-http 127.0.0.1:8082`. Replace with `-http 127.0.0.1:0` (random) or omit the flag entirely so users get the random-port default. (Grep TBD — will list affected files.)

### 2. Better startup diagnostic

In `main.go:252-256`, when `net.Listen` fails with `EADDRINUSE`, replace the cryptic `log.Fatalf` with an actionable message:

```
mcp-proxy: -http 127.0.0.1:8082 is already in use (likely another mcp-proxy session).
  - Pass -http 127.0.0.1:0 to use a random port, or
  - Pass -http=none to disable the approver entirely.
```

Don't silently rebind to a random port — a hardcoded port is a guarantee callers may rely on.

## Out of scope

OAuth token caching (the original framing of this issue) is a separate, valid feature request that is unrelated to this port collision. mcp-proxy doesn't do OAuth itself today — that lives inside the spawned `mcp-remote` child. Token caching becomes mcp-proxy's concern only when/if it owns the OAuth flow directly. Tracking separately, linked to #227 (native HTTP/SSE transport for remote MCP servers).

## Acceptance

- [ ] All `127.0.0.1:8082` references in docs/examples updated to `127.0.0.1:0` (or removed).
- [ ] `mcp-proxy` startup with a busy `-http` port prints an actionable error and exits non-zero (current behaviour: cryptic `bind: address already in use`).
- [ ] Concurrent Claude Code sessions using the random-port default both connect successfully.

ojongerius:
## Summary

- Flip default of `-http` from `127.0.0.1:0` to `none`. The HTTP approval listener now only starts when the operator explicitly passes `-http <addr>`. Pause rules still load and evaluate; an unwired pause takes the existing `ApprovalNoApprover` fail-fast path (code `-32003`) instead of the 60-second timeout.
- Replace `log.Fatalf` on bind failure with an actionable error: names the busy address and offers both `-http 127.0.0.1:0` (random) and `-http=none` (disable) as remediations. Exits non-zero.
- Soften the startup banner: default-off is now an `[INFO]` hint ("approver off by default; pass -http <addr> to enable"), not a `[WARN]`. Explicit `-http=none` stays silent. The existing WARN path remains reachable for the unusual `approverDisabled=false && approvalURL=""` case.
- Reframe docs across `claude-code.mdx`, `claude-desktop.mdx`, `codex.mdx`, `remote-servers.mdx`, `overview.mdx`, `installation.mdx`, and `README.md` so setup snippets no longer pin `-http`. The Approval Server page and Configuration reference lead with "off by default" and document the opt-in path.

## Why

Closes #262. The port collision happens because (1) the listener auto-starts whenever any `pause` rule is enabled (always true given default rules) and (2) our docs pin `-http 127.0.0.1:8082`, so two concurrent sessions both try to bind that port and the second one `log.Fatalf`s. Making the listener opt-in eliminates the collision as a category — no listener, no port, no fight — without removing the feature for users who want it.

We considered removing the `pause` action entirely (~350 LOC, no current users, no UI ships) but kept it: the marketing story (approval workflow as a first-class capability), the option to wire a bundled UI later, and the testing cost of removal all argued for preservation.

#263 (OAuth token caching for remote MCP servers) is unrelated and stays scoped against #227 — out of scope here.

## Test plan

- [x] `go vet ./... && go test ./...` from `mcp-proxy/` — passes
- [x] `mcp-proxy -h` shows new `-http` default of `none` with opt-in guidance
- [x] Default-flag boot emits `[INFO]` banner with "approver off by default" hint, no `[WARN]`
- [x] `-http 127.0.0.1:<busy-port>` produces the actionable diagnostic and exits non-zero
- [x] `cd site && pnpm build` — clean, no broken internal links to `/mcp-proxy/approval-ui/`
- [ ] Two concurrent `mcp-proxy` sessions with default flags both connect (the original #262 repro)
- [ ] Pause path with `-http 127.0.0.1:0` still functions end-to-end (listener up, approve via curl, audit row populated)

The two unchecked items need real MCP-client wiring to exercise — left for reviewer / manual smoke. Unit tests cover the banner branches, the help text, and the bind-failure message format.

### DIFF
--- a/mcp-proxy/README.md
+++ b/mcp-proxy/README.md
@@ -124,7 +124,7 @@ rules:
 
 Actions: `pass` (log only), `flag` (log + highlight), `pause` (wait for approval), `block` (reject).
 
-When a tool call is paused, approve or deny via HTTP. The approval URL and bearer token are logged to stderr at startup (the default port is random; pass `-http 127.0.0.1:PORT` to pin). Copy the token from the startup line and export it before running the curls:
+When a tool call is paused, approve or deny via HTTP. **The listener is off by default** — pass `-http 127.0.0.1:PORT` to enable it (without that flag, paused calls fail fast with JSON-RPC code `-32003`). The approval URL and bearer token are logged to stderr at startup. Copy the token from the startup line and export it before running the curls:
 
 ```sh
 export APPROVAL_TOKEN=<token-from-stderr>
--- a/mcp-proxy/cmd/mcp-proxy/main.go
+++ b/mcp-proxy/cmd/mcp-proxy/main.go
@@ -261,11 +261,7 @@ func serve() {
 	if !approverDisabled && *httpAddr != "" {
 		ln, err := net.Listen("tcp", *httpAddr)
 		if err != nil {
-			// Actionable error: name the address and offer both remediation paths.
-			fmt.Fprintf(os.Stderr,
-				"mcp-proxy: cannot bind approval listener on %s: %v\n"+
-					"  Fix: use -http 127.0.0.1:0 for a random free port, or -http=none to disable the listener.\n",
-				*httpAddr, err)
+			fmt.Fprint(os.Stderr, formatBindFailure(*httpAddr, err))
 			os.Exit(1)
 		}
 		approvalURL = "http://" + ln.Addr().String()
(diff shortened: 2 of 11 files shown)


ojongerius:
All four Copilot comments addressed in 63914de:

1. **`emitStartupBanner` doc comment claimed `-http=none` was "silent"** — rewrote the comment to describe what actually varies per config (level, suffix), since the line + JSON companion always print.
2. **`TestStartupBannerExplicitNone` comment said "completely silent"** — fixed the comment to say "suffix and WARN suppressed", and added an assertion that the JSON companion line still prints (covering the gap the original wording implied).
3. **`TestStartupBannerExplicitNone` and `TestStartupBannerNoWarnWhenApproverExplicitlyDisabled` overlapped** — deleted the older one. The new test is a strict superset (same inputs, same assertions, plus the JSON-companion check).
4. **`TestBindFailureDiagnostic` re-implemented the message** — extracted `formatBindFailure(addr, err)` in main.go and used it from both the production bind-failure path and the test. The test now also asserts on `bindErr.Error()` so a regression in either the wrapper or the underlying error surface fails the test.

Tests still green. Thanks for the review.

Issue #253, opened 2026-04-26
chore: auto-generate per-component changelogs from Conventional Commits


ojongerius:
## Background

The repo has no CHANGELOG files. Release notes are currently written by hand from commit messages (e.g. the sdk-ts v0.4.1 release). This will drift in quality over time.

The repo already enforces Conventional Commits via a pre-commit hook, which makes auto-generation straightforward.

## Proposal

Wire up [`git-cliff`](https://git-cliff.org/) with per-component path filters to generate a `CHANGELOG.md` in each SDK/component directory (`sdk/go/`, `sdk/ts/`, `sdk/py/`, `mcp-proxy/`). Changelogs would be updated as part of the release bump commit for each component.

## Tradeoffs

- **Pro:** release notes become consistent and zero-effort; no more hand-writing them
- **Pro:** existing commit history is already well-structured (Conventional Commits), so the initial generation will be meaningful
- **Con:** adds a tool dependency and some one-time setup cost
- **Con:** multi-component monorepo path filtering needs care to avoid cross-contamination between components

## Out of scope

A single root `CHANGELOG.md` — too noisy given independently versioned components.

PR #267, opened 2026-04-29
docs(mcp-proxy): update troubleshooting + add CHANGELOG

ojongerius:
## Summary

Follow-up to [#266](https://github.com/agent-receipts/ar/pull/266) (released as `mcp-proxy/v0.6.2`).

- Rewrites the troubleshooting section in `configuration.mdx` for the new opt-in default. With the listener off by default, `-32003` ("no approver configured") is the new common proxy-side failure mode and gets its own remediation block. The old "always `-32002`" framing is replaced with a three-case taxonomy (client-side denial / `-32003` / `-32002`).
- Fixes a pre-existing inaccuracy in the audit-DB diagnostic: rejected pause calls write `policy_action = 'rejected'`, not `'pause'` ([main.go:427](mcp-proxy/cmd/mcp-proxy/main.go:427)). The doc previously instructed readers to look for `policy_action = 'pause'` rows, which would never appear for a rejected call.
- Adds `approval_wait_us` to the diagnostic `SELECT` so users can distinguish `-32003` (≈0μs wait) from `-32002` (≈ `-approval-timeout` wait) post-hoc, without the error response.
- Adds `mcp-proxy/CHANGELOG.md`, mirroring [`sdk/ts/CHANGELOG.md`](sdk/ts/CHANGELOG.md). First entry documents v0.6.2. The "starts at X; earlier in git history" disclaimer matches the TS pattern and references [#253](https://github.com/agent-receipts/ar/issues/253) for the eventual auto-generation work.

## Test plan

- [x] `cd site && pnpm build` — clean, no broken internal links
- [x] CHANGELOG renders in GitHub's default Markdown viewer
- [x] Troubleshooting flow walks correctly from a `-32003` error to the right remediation block

### DIFF
--- a/mcp-proxy/CHANGELOG.md
+++ b/mcp-proxy/CHANGELOG.md
@@ -1,0 +1,37 @@
+# Changelog
+
+All notable changes to mcp-proxy are documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+This file starts at 0.6.2; earlier releases are recorded only in git history.
+A repo-wide effort to auto-generate changelogs from Conventional Commits is
+tracked in [#253](https://github.com/agent-receipts/ar/issues/253).
+
+## [0.6.2] - 2026-04-29
+
+### Changed
+
+- Default of `-http` flag flipped from `127.0.0.1:0` to `none`; the approval
+  HTTP listener is now opt-in. Pause rules still load and evaluate; an unwired
+  pause fails fast with JSON-RPC code `-32003` ("no approver configured")
+  instead of waiting 60s for a timeout. See
+  [#266](https://github.com/agent-receipts/ar/pull/266) and
+  [#262](https://github.com/agent-receipts/ar/issues/262).
+
+- Startup banner softened. The default-off case (operator did not pass `-http`)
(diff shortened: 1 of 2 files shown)


ojongerius:
All three Copilot comments addressed in 32ac3a2 — verified each against the actual code first:

1. **`created_at` doesn't exist** — confirmed: `tool_calls` schema has `requested_at` / `responded_at`, no `created_at` ([store.go:38-60](mcp-proxy/internal/audit/store.go:38)). Pre-existing bug, fixed since I touched the query.
2. **`-32002` covers both deny and timeout** — confirmed at [main.go:694-701](mcp-proxy/cmd/mcp-proxy/main.go:694) ("`-32002` covers the approved-channel cases (deny / timeout)"). Reworded case 3 as "approval denied or timed out" with `data.status` (`"denied"` / `"timed_out"`) as the disambiguator. Updated the remediation block: `"denied"` is the workflow working as intended (no proxy fix needed); `"timed_out"` is the case that needs investigation.
3. **`approval_wait_us` heuristic is unreliable** — confirmed: [`recordRejectedToolCall`](mcp-proxy/cmd/mcp-proxy/main.go:825) only sets the column when wait > 0µs, so the no-approver fail-fast path can store NULL. And a fast deny under `-32002` also has small wait, indistinguishable from `-32003` from the audit DB alone. Rewrote the bullet to recommend `data.status` first, qualifying `approval_wait_us` as a reliable signal only for the timeout sub-case.

Site builds clean.

PR #275, opened 2026-04-29
fix(sdk-ts): validate receipt schema on load from store (#170)

### LINKED ISSUE #170: sdk/ts: validate receipt schema on load from store
## Problem

`sdk/ts/src/store/store.ts:68-74` (`parseReceiptJson`) deserializes stored receipts with `JSON.parse(json) as AgentReceipt`. The `as` cast is a compile-time assertion with no runtime check. A corrupt or partial receipt in SQLite returns as a typed object with missing or wrong-typed fields and is then used throughout the SDK — chain verification, signature verification, queries.

For an audit-trail library, silent acceptance of malformed stored data is a correctness issue, not a cosmetic one.

## Comparison with other SDKs

| SDK | On-load validation | Source |
|---|---|---|
| Python | Strong — Pydantic `model_validate` | `sdk/py/src/agent_receipts/store/store.py:225` |
| Go | Moderate — `json.Unmarshal` into typed struct (catches type shape, not formats/enums) | `sdk/go/store/store.go:186` |
| TypeScript | **None** — bare `as` cast | `sdk/ts/src/store/store.ts:68-74` |

## Scope

Bring TS's on-load validation at least to Go's level, ideally to Python's. Options to evaluate — decide based on bundle size, maintenance, and consistency with the spec's JSON Schema:

- **Zod** — schema derived from existing TS types; idiomatic; adds a runtime dependency (~50KB).
- **Hand-written minimal validator** — no new dependencies; more maintenance as the schema evolves.
- **ajv + `spec/schema/agent-receipt.schema.json`** — uses the JSON Schema that already exists in the repo; stronger consistency story with Python/Go; adds a heavier dependency.

The SDK currently has **zero runtime dependencies**. Adding one warrants an ADR or at least an explicit note in the PR description.

## Related

`signReceipt`, `verifyReceipt`, and `createReceipt` also skip semantic validation. Decide whether this issue covers the full API surface or scopes to store-load only. Recommend **store-load first**; follow up separately on create/sign/verify if warranted.

## Acceptance

- Loading a deliberately corrupted receipt JSON from the store surfaces a clear, contextual error (not a silent ghost-typed object).
- Test covers the failure mode.
- If a runtime dependency is added, it's documented (ADR or PR description justifying the choice).

---
_Source: internal code-review pass, April 2026._

ojongerius:
Closes #170.

## Summary

- `parseReceiptJson` now validates every receipt loaded from SQLite against a Zod schema mirroring the `AgentReceipt` interface, instead of relying on a compile-time-only `as AgentReceipt` cast that silently passed corrupt rows through to chain/signature verification.
- Errors keep the existing `Corrupt receipt in store (<context>)` shape and now include the offending field path (e.g. `credentialSubject.chain.chain_id: Required`).
- Brings TS on-load validation past Go's level (typed `json.Unmarshal`) toward Python's Pydantic level. Scope is store-load only; create/sign/verify validation is explicitly deferred per the issue.

## New runtime dependency: `zod`

This is the SDK's first runtime dependency. Rationale captured in [ADR-0011](docs/adr/0011-zod-for-runtime-validation.md):

- **Hand-written validator** — drifts from `types.ts`, no compile-time guarantee they stay in sync.
- **ajv + JSON Schema** — heavier dep; JSON Schema and TS types have different evolution cadences.
- **Zod** — schemas live next to TS types; `z.infer` aligns the two at compile time; ecosystem standard.

Adds ~50 KB to install size; tree-shaking applies for bundled consumers.

## Notable design choices

- Schemas are **not** `.strict()` — unknown extra fields written by a newer SDK do not break older readers.
- `chain.terminal` is `z.literal(true).optional()` so a stored `false` value (spec-invalid per [types.ts:124-126](sdk/ts/src/receipt/types.ts:124)) is rejected at load time.
- `version` stays `z.string()` (not an enum) so older receipts still load.
- The existing `AgentReceipt` interface is unchanged; the schema's inferred type is structurally compatible but separate, to avoid churning callers.

## Files

- `sdk/ts/package.json` — add `zod` to `dependencies`.
- `sdk/ts/src/receipt/schema.ts` — new Zod schema mirroring `types.ts`.
- `sdk/ts/src/store/store.ts` — `parseReceiptJson` now runs `agentReceiptSchema.parse` after `JSON.parse` and re-throws `ZodError` with field paths in the existing message format.
- `sdk/ts/src/store/store.test.ts` — 6 new tests across all three caller paths.
- `docs/adr/0011-zod-for-runtime-validation.md` — new ADR.

## Test plan

- [x] `pnpm test` — 183 tests pass (was 177; +6 new).
- [x] `pnpm build` — `tsc` clean.
- [x] `pnpm exec biome check src` — clean.
- [x] New tests cover: non-JSON garbage, missing required field, invalid enum, `chain.terminal: false`, plus per-caller coverage for `getById`, `getChain`, and `query`.

### DIFF
--- a/docs/adr/0011-zod-for-runtime-validation.md
+++ b/docs/adr/0011-zod-for-runtime-validation.md
@@ -1,0 +1,60 @@
+# ADR-0011: Zod for runtime schema validation
+
+## Status
+
+Accepted
+
+## Context
+
+`parseReceiptJson` in the TypeScript SDK deserialised stored receipts with a
+bare `JSON.parse(json) as AgentReceipt` cast. The cast is compile-time only —
+a corrupt or partially-written SQLite row returns as a ghost-typed `AgentReceipt`
+with missing or wrong-typed fields, which then flows silently into chain
+verification, signature verification, and queries.
+
+For an audit-trail library this is a correctness gap, not a cosmetic one.
+The Python SDK validates via Pydantic `model_validate`; the Go SDK relies on
+`json.Unmarshal` against a typed struct. The TypeScript SDK had no equivalent.
+
+Issue #170 asked for runtime validation at the store-load boundary. Three
+options were evaluated:
+
+1. **Hand-written validator** — covers the need but is maintenance-heavy: every
+   field added to the spec requires a parallel update in the validator, and
+   there is no type-level guarantee that the two stay in sync.
+
+2. **ajv + JSON Schema** — the spec ships JSON schemas, so this would be
+   authoritative. However, ajv is a significantly larger dependency, JSON Schema
+   and TypeScript types have different evolution cadences, and keeping both in
+   sync would require additional tooling.
+
+3. **Zod** — schemas are written in TypeScript alongside the types, `z.infer`
+   ensures structural alignment at compile time, and the library is the de-facto
+   standard for runtime validation in the TypeScript ecosystem (>30M weekly
+   downloads, actively maintained).
+
+## Decision
+
+Adopt Zod for runtime validation at the store-load boundary. A Zod schema
+mirroring every interface in `src/receipt/types.ts` is defined in
+`src/receipt/schema.ts` and called inside `parseReceiptJson` after
+`JSON.parse`. Schema objects are not strict (no `.strict()`) so unknown extra
+fields introduced by newer SDK versions do not break older stores. The existing
+`AgentReceipt` interface is kept unchanged; the schema type is structurally
+compatible but separate.
+
+This is the SDK's first runtime dependency.
+
+## Consequences
+
+- Corrupt or schema-invalid rows in the store now surface a clear, field-pathed
+  error at load time rather than propagating ghost-typed values downstream.
+- Zod adds approximately 50 KB to the installed package (tree-shaking applies
+  for bundled consumers).
+- Setting a precedent for a runtime dependency: future additions must clear the
+  same bar (well-maintained, widely adopted, supply-chain vetted).
+- Validation in `createReceipt`, `signReceipt`, and `verifyReceipt` is
+  explicitly deferred; the same schema can be reused at those sites if issue
+  #170 scope is expanded later.
+- If cross-SDK schema consistency becomes a priority (ajv + canonical JSON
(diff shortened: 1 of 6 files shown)


ojongerius:
Addressed Copilot's review in [b273e50](https://github.com/agent-receipts/ar/pull/275/commits/b273e50):

- **Unknown-field stripping (correctness)** — every `z.object` now uses `.passthrough()`. The original schemas would have silently dropped unknown keys, invalidating the RFC 8785 canonical hash on receipts written by newer SDK versions and breaking signature verification. ADR-0011 updated to describe the decision accurately.
- **Forward-compat regression test** — new test inserts a receipt with extra unknown top-level and nested fields and asserts they survive the load. Would have caught the bug above.
- **`as AgentReceipt` cast at the validation boundary** — removed. The exported schema is now annotated as `z.ZodType<AgentReceipt>` so `.parse()` returns `AgentReceipt` directly.
- **Test helper double-cast** — replaced with `// @ts-expect-error` access on the private `db` field; a future rename will now fail compilation instead of silently going untyped.
- **First runtime dep flag** — already covered by [ADR-0011](docs/adr/0011-zod-for-runtime-validation.md) and the PR description; no code change needed.

Tests: 184 pass (was 183, +1 forward-compat). `tsc` clean, biome clean.

ojongerius:
Addressed Copilot's follow-up in [01f5745](https://github.com/agent-receipts/ar/pull/275/commits/01f5745):

- **`Error.cause` on both rethrow paths** — JSON.parse failure now wraps the original `SyntaxError` as `cause`; schema failure wraps the `ZodError` (so callers can inspect `.cause.issues`).
- **Root-path rendering** — when a Zod issue has `path = []` (e.g. the stored value is a JSON primitive), the message now reads `... (root): Expected object, received number` instead of `... : Expected object, received number`.
- **First runtime dep flag** — repeat from the previous round; covered by [ADR-0011](docs/adr/0011-zod-for-runtime-validation.md). No code change.

Three new regression tests cover cause propagation on both branches and the `(root)` rendering. Tests: 187 pass (was 184, +3). tsc clean, biome clean.

Issue #280, opened 2026-04-29
Promote parameterDisclosure to first-class feature across all SDKs and MCP proxy


ojongerius:
Currently `parameterDisclosure` (full payload storage vs hash-only) is implemented in the OpenClaw plugin under the legacy name `parameterPreview`, but is not surfaced as a first-class feature across the MCP proxy sidecar and TypeScript / Python / Go SDKs.

This issue tracks both the cross-channel rollout and the rename from `parameterPreview` → `parameterDisclosure` (config knob) and `parameters_preview` → `parameters_disclosure` (TS SDK field). The new name survives the design pivot in ADR-0012: the field carries asymmetric ciphertext, not plaintext, so "preview" / "raw" / "plain" all mis-describe it.

## Why this matters

Privacy-preserving by default (hash-only), full forensics on demand (asymmetric ciphertext, recoverable only with the forensic private key) is a strong design position that directly answers the most common security review question: "how do I know what was actually sent in case of a breach?" The asymmetric construction means *the emitter cannot read its own past disclosures* — same trust property that ADR-0010 establishes for the audit trail itself.

## Requirements

- Operator-level config only — an agent must never be able to disable its own payload logging
- Consistent behaviour across all SDKs and MCP proxy
- Asymmetric encryption: emitter holds public key only; private key lives with forensic responder / escrow
- Forward-compatible envelope shape — single-recipient v1 must extend to multi-recipient later without breaking older verifiers
- Prominently documented as a conscious design decision, not buried in installation config

## Risks to address in implementation

- Even encrypted-in-body, the forensic private key becomes a high-value asset — rotation, backup, escrow, and recovery stories needed
- Tension with daemon separation goal (keeping secrets out of agent process vs storing them in daemon store) — daemon mode resolves this; SDK-direct mode keeps the plaintext window in the SDK process
- GDPR / privacy implications even with crypto-shredding (key destruction → ciphertext meaningless) — operator responsibility, needs clear documentation
- Cross-SDK canonical-JSON shape of the encrypted disclosure field becomes load-bearing (see ADR-0009)
- Receipts are immutable, so old forensic private keys must be retained forever to decrypt historical disclosures

## Mode ladder

The same architecture serves three personas with no migration between them:

- **Solo dev** — auto-generated forensic key pair on first run, both halves local. Out-of-the-box visibility.
- **Small team** — public key distributed to laptops; private key held by security lead. No per-laptop secret distribution.
[...]

PR #281, opened 2026-04-29
docs(adr): draft ADR-0012 for payload disclosure policy

ojongerius:
Drafts ADR-0012 covering the design decision for first-class `parameterDisclosure` across all SDKs and the MCP proxy. Tracks implementation under #280.

## What this PR contains

- **`docs/adr/0012-payload-storage-policy.md`** — new Draft ADR.
- **`docs/adr/README.md`** — adds ADR-0012 row. Also adds the previously-missing ADR-0011 row.

No code changes. The ADR is a scaffold for discussion alongside #280; the implementation work lives there in three sequenced phases.

## What ADR-0012 commits to

- **Hash-only by default.** Identical commitment via `parameters_hash` on every channel.
- **Opt-in disclosure** via an operator-only `parameterDisclosure` config knob. Agents cannot self-enable or self-disable.
- **Asymmetric encryption** of the disclosure field inside the signed receipt body. The emitter holds only the public key; the private key lives with the forensic responder / escrow holder. The encryptor genuinely cannot read its own past output — same trust property as ADR-0010.
- **Forward-compatible envelope** — `{ v, alg, recipients[], nonce, ct }`. Single-recipient v1 extends to multi-recipient HPKE later without breaking older verifiers.
- **Storage adapters stay opaque.** SQLite today, Postgres / S3 / OTel exporters tomorrow — none of them see plaintext or inherit key-management responsibility.
- **Daemon owns fan-out** (per ADR-0010). The same signed receipt goes to local store, SIEM, and archive sinks; the forensic private key gates disclosure mathematically, not configurationally.
- **Crypto-shredding** is the GDPR right-to-erasure story.

## Three-mode ladder (one architecture)

| | Solo dev | Small team | Enterprise |
|---|---|---|---|
| Public key | Auto-generated locally | Distributed via config or daemon | Daemon-served from registry |
| Private key | Local | Held by security lead | HSM / KMS, multi-recipient escrow |
| Storage | Local SQLite | Local + optional sync | Pluggable remote |
| Fan-out | None | Optional SIEM | SIEM + archive + compliance |

A solo dev growing up does not re-encrypt receipts; only key custody changes.

## Naming

- Config knob: `parameterDisclosure` (renamed from `parameterPreview`).
- Receipt field: `parameters_disclosure` (renamed from the existing TS `parameters_preview`).

The previous name described plaintext; ciphertext-in-body required something that survived the semantic shift. "Disclosure" reads honestly to a forensic reader (with the key) and to a SIEM operator (without).

## Why now, not after the daemon refactor (ADR-0010)

The receipt format is permanent once signed. Settling the envelope shape and the asymmetric model now means Phase A receipts (SDK-direct, solo-dev) are forward-compatible with Phase B (post-daemon) and Phase C (enterprise) without re-encryption. Implementation phasing follows the daemon timeline; the format does not wait.

## Notes for reviewers

- Status is `Draft` — this is an extension of the template's `Proposed | Accepted | Deprecated | Superseded`. Happy to swap to `Proposed` if strict template adherence is preferred.
- Cross-SDK serialisation of the disclosure envelope is load-bearing (ADR-0009 territory). The ADR commits to the shape but defers the exact `alg` string and `kid` registry mechanism to a follow-up implementation ADR.

Refs: #280
Related: ADR-0009, ADR-0010, ADR-0011

### DIFF
--- a/docs/adr/0012-payload-disclosure-policy.md
+++ b/docs/adr/0012-payload-disclosure-policy.md
@@ -2,13 +2,13 @@
 
 ## Status
 
-Draft
+Proposed
 
 ## Context
 
 Receipts today commit to action parameters via `parameters_hash` only. That is the right default — it is privacy-preserving, tamper-evident, and small. But the most common forensic question after an incident is "what did the agent actually send?" and a hash cannot answer it.
 
-The OpenClaw plugin already documents an opt-in `parameterPreview` config that selectively discloses parameters by risk class. The TypeScript SDK exposes the field shape as `parameters_preview` on `Action` ([sdk/ts/src/receipt/types.ts:59](../../sdk/ts/src/receipt/types.ts)), with a CHANGELOG warning that the value is permanent and signed and "must never be auto-populated from raw arguments". Python and Go SDKs do not have the field. The MCP proxy has no equivalent knob, though it does have opt-in AES-256-GCM encryption of redacted audit fields via `BEACON_ENCRYPTION_KEY` ([mcp-proxy/cmd/mcp-proxy/main.go:204](../../mcp-proxy/cmd/mcp-proxy/main.go)) — prior art for non-signing key handling.
+The OpenClaw plugin already documents an opt-in `parameterPreview` config that selectively discloses parameters by risk class. The TypeScript SDK exposes the field shape as `parameters_preview` on `Action` ([sdk/ts/src/receipt/types.ts](../../sdk/ts/src/receipt/types.ts)), with a CHANGELOG warning that the value is permanent and signed and "must never be auto-populated from raw arguments". Python and Go SDKs do not have the field. The MCP proxy has no equivalent knob, though it does have opt-in AES-256-GCM encryption of redacted audit fields via `BEACON_ENCRYPTION_KEY` ([mcp-proxy/cmd/mcp-proxy/main.go](../../mcp-proxy/cmd/mcp-proxy/main.go)) — prior art for non-signing key handling.
 
 The result is that the forensic question gets a different answer depending on which channel produced the receipt. We want a uniform, operator-controlled, privacy-preserving-by-default position across every emitter, documented as a deliberate design decision rather than buried in installation config.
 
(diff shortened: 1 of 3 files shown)


ojongerius:
Wrap-up note for the audit trail — all 9 Copilot inline comments are addressed in the merged commits:

**Round 1 (commit `1d7c6f0`):**
- [r3160379581](https://github.com/agent-receipts/ar/pull/281#discussion_r3160379581) — `types.ts:59` line number dropped (was wrong; field is at line 76 anyway).
- [r3160379654](https://github.com/agent-receipts/ar/pull/281#discussion_r3160379654) — ADR Status `Draft` → `Proposed` (matches template vocabulary).
- [r3160379700](https://github.com/agent-receipts/ar/pull/281#discussion_r3160379700) — same Status fix on the README index row.
- [r3160379747](https://github.com/agent-receipts/ar/pull/281#discussion_r3160379747) — `main.go:204` line number dropped (was the taxonomy block; `BEACON_ENCRYPTION_KEY` is actually around line 231).
- [r3160379780](https://github.com/agent-receipts/ar/pull/281#discussion_r3160379780) — file renamed to `0012-payload-disclosure-policy.md` so the slug matches the title (the ADR is specifically about disclosure, not storage in general).

**Round 2 (commit `c87e1f5`):**
- [r3165864361](https://github.com/agent-receipts/ar/pull/281#discussion_r3165864361) — Context paragraph rewritten to distinguish whole-store at-rest encryption (which is fine; doesn't change signed bytes) from per-sink selective encryption (which is the real conflict with plaintext-in-body).
- [r3166112119](https://github.com/agent-receipts/ar/pull/281#discussion_r3166112119) — CHANGELOG paraphrase aligned with the actual wording (`MUST NOT populate parameters_preview from raw tool arguments`); quotation marks dropped.
- [r3166112144](https://github.com/agent-receipts/ar/pull/281#discussion_r3166112144) — invented `agent-receipts inspect` replaced with the existing `mcp-proxy inspect <id>` subcommand ([cli.go:234](https://github.com/agent-receipts/ar/blob/main/mcp-proxy/cmd/mcp-proxy/cli.go#L234)).

**Final round (post-merge, [r3166425032](https://github.com/agent-receipts/ar/pull/281#discussion_r3166425032)):**
- Fair catch — the PR description was not kept in sync with the file rename and the `Draft` → `Proposed` status change. The merged ADR is the source of truth; the PR body is preserved as-is for historical context. Future ADR PRs should refresh the description after rename / status edits.

Implementation work continues in #280.

PR #282, opened 2026-04-29
feat(mcp-proxy): expand secret redaction patterns and add audit-secrets scan

### LINKED ISSUE #150: Audit secret redaction patterns for edge cases
## Problem

The proxy redacts GitHub PATs and other secrets before storing receipts. However, tokens can appear in unexpected places, and the current redaction patterns may not cover all cases:

- **URL parameters** — e.g. `?token=ghp_...` or `?access_token=...` in API responses or error messages
- **Error messages** — GitHub API error responses sometimes echo back auth headers or tokens
- **Non-standard token formats** — fine-grained PATs (`github_pat_...`), OAuth tokens, app installation tokens (`ghs_...`, `ghu_...`)
- **Base64-encoded tokens** — tokens embedded in base64 blobs (e.g. in JWT payloads)
- **Multi-line values** — tokens split across JSON fields or spanning line boundaries

A missed redaction means a secret ends up in the receipt DB permanently — there's no retroactive redaction mechanism.

## Proposal

- Audit the current redaction regex patterns against all known GitHub token formats
- Add test cases for each edge-case location listed above
- Support a configurable redaction pattern list (`-redact-patterns` flag or config file) for custom secrets (e.g. Slack tokens, API keys for other services)
- Consider a catch-all high-entropy string detector as a safety net

## Acceptance criteria

- [ ] All GitHub token formats are covered: classic PATs (`ghp_`), fine-grained (`github_pat_`), OAuth (`gho_`), app tokens (`ghs_`, `ghu_`), installation tokens (`v1.`)
- [ ] Tokens in URL parameters (`?token=`, `?access_token=`) are redacted
- [ ] Tokens echoed in API error response bodies are redacted
- [ ] Test suite includes a test case for each token format and each location (header, URL param, response body, error message)
- [ ] Configurable custom redaction patterns via flag or config file, documented in docs
- [ ] Existing receipts with leaked secrets can be identified (e.g. `mcp-proxy audit-secrets` scan command)
- [ ] Docs updated with: list of redacted patterns, how to add custom patterns, what to do if a secret is found in the receipt DB

ojongerius:
Closes #150.

## Summary

- Adds patterns for fine-grained PATs (`github_pat_`), `ghs_`, `ghu_`, legacy `v1.<hex>` installation tokens, and URL-parameter tokens (`?token=`, `?access_token=`, `?api_key=`, `?auth=`, mixed case). The URL-param replacement preserves the key name so debugging context survives.
- Introduces a `Redactor` struct + `NewRedactor(custom []*regexp.Regexp)` so callers can layer custom patterns on top of the built-ins without mutating package state. The package-level `Redact(raw)` shim is preserved so existing call sites are untouched.
- Adds `LoadPatterns` (YAML) and a `-redact-patterns` flag on `serve`. Example file at `mcp-proxy/configs/example_redact_patterns.yaml`.
- Adds `mcp-proxy audit-secrets` subcommand that scans `messages.raw` and `tool_calls.{arguments,result,error}` for unredacted secrets the live redactor missed. Runs both the regex pattern set **and** a JSON-key pass, mirroring the live redactor's two-pass logic. Honours `BEACON_ENCRYPTION_KEY` for encrypted DBs; decrypt failures are reported as hits and the scan continues. Exit 0 = clean, 1 = hits, 2 = operational error. Output never echoes the matched secret.
- Tests cover each token format × each location (header, URL param, JSON body, plain error string, multi-line PEM), case-insensitive URL params, custom patterns, `ScanJSONLeaks`, and the `audit-secrets` exit-code paths including the encrypted-DB decrypt-error path.

## Notes / deviations from the issue

- The proposed `-key-file` flag is replaced by the existing `BEACON_ENCRYPTION_KEY` env var (the proxy already uses this for encryption-at-rest). Documented in the README.
- The high-entropy catch-all detector mentioned in the issue is intentionally out of scope here — false-positive risk on commit hashes / UUIDs / opaque IDs warrants a separate discussion. Will file a follow-up issue if useful.
- `audit-secrets` does not open the receipt store — receipts contain hashes, not bodies, so there is nothing leak-scannable there. Only the audit DB needs scanning.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -count=1 ./...` (all 5 mcp-proxy packages green)
- [x] Smoke: `mcp-proxy audit-secrets -db <empty>.db` → exit 0
- [ ] Reviewer: try `mcp-proxy serve -redact-patterns configs/example_redact_patterns.yaml ...`, send a tool call with a Slack token, confirm the stored row shows `[REDACTED]`
- [ ] Reviewer: seed an audit DB with a known leaked token, run `mcp-proxy audit-secrets -db <path>`, confirm the row is reported and exit code is 1

### DIFF
--- a/mcp-proxy/README.md
+++ b/mcp-proxy/README.md
@@ -172,6 +172,72 @@ Paused calls auto-deny after 60 seconds (fail-safe).
 
 Set `BEACON_ENCRYPTION_KEY` to enable AES-256-GCM encryption at rest for sensitive audit data.
 
+## Secret redaction
+
+The proxy redacts secrets before writing to the audit database in two passes:
+1. **JSON-key pass** — any value whose JSON key matches a sensitive name (e.g. `password`, `token`, `api_key`) is replaced with `[REDACTED]`.
+2. **Pattern pass** — 12 built-in regular expressions catch common token formats regardless of key name.
+
+Built-in patterns:
+
+| Name | Matches |
+|------|---------|
+| `github-pat-classic` | `ghp_…` GitHub personal access tokens |
+| `github-pat-finegrained` | `github_pat_…` fine-grained PATs |
+| `github-oauth` | `gho_…` OAuth tokens |
+| `github-app-installation` | `ghs_…` GitHub App installation tokens |
+| `github-user-to-server` | `ghu_…` user-to-server tokens |
+| `github-installation-legacy` | `v1.<40+ hex chars>` legacy installation tokens |
+| `openai-anthropic-key` | `sk-…` OpenAI/Anthropic secret keys |
+| `aws-access-key` | `AKIA…` AWS access key IDs |
+| `bearer-token` | `Bearer <token>` HTTP Authorization headers |
+| `slack-token` | `xoxb-/xoxp-/xoxr-/xoxa-/xoxs-` Slack tokens |
+| `pem-private-key` | PEM `-----BEGIN … PRIVATE KEY-----` blocks |
+| `url-param-token` | `?token=…`, `?access_token=…`, `?key=…` etc. (key name preserved) |
+
+### Custom patterns
+
+Add organisation-specific patterns with a YAML file:
+
+```yaml
+# custom_redact.yaml
+patterns:
+  - name: slack-webhook
+    pattern: 'https://hooks\.slack\.com/services/[A-Z0-9/]+'
+  - name: stripe-live
+    pattern: 'sk_live_[A-Za-z0-9]{24,}'
+```
+
+Pass it at startup:
+
+```sh
+mcp-proxy -redact-patterns custom_redact.yaml -- npx -y @modelcontextprotocol/server-filesystem /
+```
+
+An example file is at `configs/example_redact_patterns.yaml`.
+
+### Auditing existing databases
+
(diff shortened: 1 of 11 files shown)


ojongerius (commented):
Round 3 review addressed in e825d22.

1. **`BuiltinPatterns()` shared `*regexp.Regexp` pointers** — fixed in `mcp-proxy/internal/audit/redact.go` by recompiling each `Re` from its source string inside the loop. Callers can now safely call mutating methods like `Regexp.Longest()` on a returned pattern without affecting package state, matching what the docstring claims.

2. **`OpenReadOnly` DSN percent-encoding** — fixed in `mcp-proxy/internal/audit/store.go`. The DSN is now built with `url.URL{Path: filepath.ToSlash(dbPath)}` so spaces, `?`, `#`, etc. are percent-encoded correctly. `:memory:` keeps the `Opaque` form (Path would prepend a `/`). Added `TestOpenReadOnlyPathWithSpaces` as a regression guard.

3. **`LoadPatterns` accepting empty pattern strings** — fixed in `mcp-proxy/internal/audit/redact_config.go`. After the name check, we now reject `strings.TrimSpace(p.Pattern) == ""` with `pattern %q: pattern is required`. Added `TestLoadPatternsEmptyPattern` and `TestLoadPatternsWhitespacePattern`.

All five mcp-proxy packages still pass (`go build ./...`, `go vet ./...`, `go test -count=1 ./...` all exit 0).

PR #285, opened 2026-04-30
sdk/ts + openclaw: rename parameters_preview → parameters_disclosure

ojongerius:
Closes #283 (in part — TS SDK + OpenClaw docs portion).

Per [ADR-0012](https://github.com/agent-receipts/ar/blob/main/docs/adr/0012-payload-disclosure-policy.md), the field name `parameters_preview` misdescribed a permanent, signed value: receipts are durable, so anything written there is a deliberate disclosure rather than a transient preview. This PR renames the field on the TypeScript surface and updates the OpenClaw plugin docs in lockstep. The receipt shape is unchanged (`Record<string, string>`); only the key moved. The encrypted-envelope half of ADR-0012 is a future follow-up.

**Merge order: 2 of 4** — depends on #254 (spec rename), unblocks #255 (Go SDK) and #256 (Python SDK).

### TypeScript SDK (`sdk/ts/`)

- `src/receipt/types.ts` — `Action.parameters_preview` → `parameters_disclosure`; JSDoc rewritten and an ADR-0012 migration note appended.
- `src/receipt/schema.ts` — Zod field renamed in the action schema.
- `src/receipt/types.test.ts` — full-receipt fixture uses the new key.
- `src/receipt/create.test.ts` — pass-through test renamed and updated.
- `package.json` — version bumped 0.5.0 → 0.6.0 (breaking rename; pre-1.0 protocol stays at 0.1.0).
- `CHANGELOG.md` — new [0.6.0] entry under \"Changed\" with a BREAKING marker and migration note.

### OpenClaw plugin docs (`site/src/content/docs/openclaw/`)

- `installation.mdx` — config key `parameterPreview` → `parameterDisclosure`; example output uses `parameters_disclosure`; surrounding prose updated; appended a backwards-incompat note pointing at ADR-0012.

> The OpenClaw plugin source lives in a sibling repo (`agent-receipts/openclaw`), not in this repo, so the runtime config-key rename for the plugin itself will need a matching PR there. This PR covers only what lives in `agent-receipts/ar`.

### Out of scope

- `cross-sdk-tests/` fixtures — separate follow-up.
- `spec/` schemas — handled in #254.
- Go and Python SDKs — handled in #255 and #256.

## Breaking change

[...]

PR #287, opened 2026-04-30
feat(sdk-go): add parameters_disclosure to Action

### LINKED ISSUE #255: sdk/go: add parameters_preview to Action
## Background

`@agnt-rcpt/sdk-ts` v0.5.0 adds an optional `parameters_preview` field to its `Action` interface. The Go SDK should mirror it for cross-SDK parity. Until it does, Go-side decoding silently drops the field on round-trip — signatures still verify (the field is part of the signed payload), but Go code cannot read or set the preview.

## What needs to change

Add the field to the `Action` struct in [`sdk/go/receipt/types.go`](https://github.com/agent-receipts/ar/blob/main/sdk/go/receipt/types.go) (around lines 69-78):

```go
type Action struct {
    ID                 string            `json:"id"`
    Type               string            `json:"type"`
    ToolName           string            `json:"tool_name,omitempty"`
    RiskLevel          RiskLevel         `json:"risk_level"`
    Target             *ActionTarget     `json:"target,omitempty"`
    ParametersHash     string            `json:"parameters_hash,omitempty"`
    ParametersPreview  map[string]string `json:"parameters_preview,omitempty"` // new
    Timestamp          string            `json:"timestamp"`
    TrustedTimestamp   string            `json:"trusted_timestamp,omitempty"`
}
```

## Semantics

- Optional (`omitempty`), additive.
- Flat `map[string]string` — never nested, never non-string values.
- Operator-controlled: SDK never auto-populates. Callers are responsible for ensuring only allowlisted fields appear (per the taxonomy `preview_fields` config).
- The existing `ParametersHash` still covers the full parameter set; `ParametersPreview` is display/audit metadata only.

## Acceptance

- [ ] Field added with correct JSON tag.
- [ ] Test coverage: round-trip a receipt with `parameters_preview` set; assert it survives marshal/unmarshal and is included in the canonical hash.
- [ ] `go test ./...` passes for sdk/go and mcp-proxy (workspace consumer).
- [ ] Released as a minor version bump (current is the equivalent of sdk-ts pre-0.5.0).

## Related

- TS SDK: shipped in v0.5.0.
- Spec schema update: tracked separately.
- Python SDK: tracked separately.

ojongerius:
Closes #255. Per [ADR-0012](https://github.com/agent-receipts/ar/blob/main/docs/adr/0012-payload-disclosure-policy.md).

**Merge order: 3 of 4** — depends on #286 (spec) and #285 (TS rename); unblocks the Py PR.

## Summary

Adds an optional `ParametersDisclosure map[string]string` field to the `Action` struct in `sdk/go/receipt/types.go`. The Go SDK has not shipped a prior name, so this is a clean addition.

The encrypted envelope shape from ADR-0012 is deferred to a follow-up; this change covers the simple string-map form only.

## Tests

`sdk/go/receipt/parameters_disclosure_test.go` adds three round-trip tests:

- Marshal/unmarshal preserves all entries.
- Nil and empty maps are omitted via `omitempty` (existing receipts canonicalize identically).
- Field appears in RFC 8785 canonical form, and changing it changes the receipt hash.

## Verification

- [x] `go test ./...` passes in `sdk/go/`
- [x] `go test ./...` passes in `mcp-proxy/` (workspace consumer)
- [x] `go vet ./...` clean

## Scope

- No spec, TS, or Py code touched.
- No `cross-sdk-tests/` fixtures touched (deferred to a follow-up PR after all four SDKs land).

### DIFF
--- a/sdk/go/receipt/parameters_disclosure_test.go
+++ b/sdk/go/receipt/parameters_disclosure_test.go
@@ -1,0 +1,154 @@
+package receipt
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+)
+
+// TestParametersDisclosureRoundTrip verifies that the optional
+// parameters_disclosure map (ADR-0012) survives JSON marshal/unmarshal,
+// is included in the canonical JSON form, and therefore contributes to
+// the receipt hash.
+func TestParametersDisclosureRoundTrip(t *testing.T) {
+	disclosure := map[string]string{
+		"path":   "/etc/hosts",
+		"flags":  "ro",
+		"reason": "diagnostic read",
+	}
+
+	action := Action{
+		ID:                   "act_test",
+		Type:                 "filesystem.file.read",
+		RiskLevel:            RiskLow,
+		ParametersHash:       "sha256:deadbeef",
+		ParametersDisclosure: disclosure,
+		Timestamp:            "2026-04-28T00:00:00Z",
+	}
+
+	// Marshal the Action and confirm parameters_disclosure is present.
+	encoded, err := json.Marshal(action)
+	if err != nil {
+		t.Fatalf("marshal action: %v", err)
+	}
+	if !strings.Contains(string(encoded), `"parameters_disclosure"`) {
+		t.Fatalf("expected parameters_disclosure in encoded JSON, got %s", encoded)
+	}
+
+	// Unmarshal and ensure the field round-trips with all entries intact.
+	var decoded Action
+	if err := json.Unmarshal(encoded, &decoded); err != nil {
+		t.Fatalf("unmarshal action: %v", err)
+	}
+	if len(decoded.ParametersDisclosure) != len(disclosure) {
+		t.Fatalf("expected %d disclosure entries, got %d",
+			len(disclosure), len(decoded.ParametersDisclosure))
+	}
+	for k, v := range disclosure {
+		if got := decoded.ParametersDisclosure[k]; got != v {
+			t.Errorf("disclosure[%q] = %q, want %q", k, got, v)
+		}
+	}
+}
+
+// TestParametersDisclosureOmitEmpty verifies the field is omitted from
+// JSON when nil or empty (omitempty), so existing receipts without
+// disclosure data continue to canonicalize identically.
+func TestParametersDisclosureOmitEmpty(t *testing.T) {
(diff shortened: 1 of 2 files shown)


ojongerius (commented):
## Verdict: LGTM

No blockers. The change matches #255 exactly: `ParametersDisclosure map[string]string` placed between `ParametersHash` and `Timestamp`, JSON tag `parameters_disclosure,omitempty`. Scope is tight — only `sdk/go/receipt/types.go` modified and one new test file added. All other Go source SHAs (store, taxonomy, cross_language_test.go, hash.go, signing.go, create.go, chain.go, etc.) are byte-identical to main. No mcp-proxy edits, which is correct — `Action` literals there should not auto-populate this field per ADR-0012's operator-control rule.

### Notes

- Tests cover all three acceptance criteria via real `Create`/`Sign`/`HashReceipt`/`Canonicalize`/`GenerateKeyPair` (no stubs):
  - `TestParametersDisclosureRoundTrip` — JSON marshal/unmarshal preserves the map.
  - `TestParametersDisclosureOmitEmpty` — both `nil` and `map[string]string{}` are omitted from JSON, which is load-bearing for cross-SDK hash equivalence (a Go caller doing `make(map[string]string)` won't sneak `"parameters_disclosure":{}` into the canonical form).
  - `TestParametersDisclosureCanonicalIncluded` — proves the field flows into the canonical form and changes the receipt hash; uses fixed IDs/timestamps to make the substring assertion deterministic.
- Canonicalization path is fine. `Canonicalize` re-marshals via `marshalNoHTMLEscape`, unmarshals to `map[string]any`, and `canonicalizeValue` performs RFC 8785 UTF-16 key sorting — so the new map's keys are sorted by the canonicalizer, not by Go's `encoding/json` default. Same path all other maps already use.
- Wire name `parameters_disclosure` matches ADR-0012 and the sibling PRs (#285, #286, #288).
- `Version` constant unchanged at `0.2.0` (protocol version, separate from any package release tag). The "minor version bump" mentioned in #255 is a release-time concern; flagging only so it isn't forgotten when tagging.
- `TestParametersDisclosureCanonicalIncluded` asserts a single-entry disclosure substring; if a future change adds multi-entry assertions in this style, the keys must appear in UTF-16 sort order in the canonical form. Non-blocking.

Issue #291, opened 2026-04-30
Site action plan: track positioning / accuracy cleanup across docs


ojongerius:
Tracking issue for the site audit cleanup. The plan itself lives at [`docs/audits/site-action-plan.md`](https://github.com/agent-receipts/ar/blob/4d163ad62246fcdb485812eaa3e3d2d2ea9141d3/docs/audits/site-action-plan.md) (15 blockers, 35 drift, 8 nits, 15 diagram gaps across 24 pages).

**How to use this issue:** check items off as PRs land, link the PR next to the box. Tracks 1 and 2 run in parallel; Track 3 is gated on Track 1 decisions; Track 5 is gated on the architecture/disclosure framing chosen in Track 1.

**Status (2026-05-15):** All tracks complete. All 15 blockers, 35 drift items, 8 nits, and 15 diagram gaps addressed across 12 PRs. Closing.

---

## Track 1 — Cross-cutting decisions (gating)

These are policy calls, not copy edits. Resolving them collapses ~half the per-page drift.

- [x] **CC-1** Enforcement vs. audit-only — **Option C: split story.** Protocol is audit-only; mcp-proxy `block` / `pause` / Approval Server reframed as proxy-layer features (not Agent Receipts protocol features). Pitch.md amendment + Track 3 page rewrites pending; ground-truth update lands first.
- [x] **CC-2** Daemon-isolated keys — **Dissolved.** ADR-0010 (#236) ships before any Track 3 docs work, so docs match the v2 daemon architecture in one pass instead of writing v1 framing that gets rewritten.
- [x] **CC-3** ADR-0010 disclosure — **Dissolved** (same reason as CC-2).
- [x] **CC-4** Specification SVG signing-attribution — **Dissolved** (same reason).
- [x] **CC-5** Hero block — **Adopt `pitch.md` H1 + subhead verbatim** on `index.mdx`. Sequenced after ADR-0010 ships so the daemon claim is honest at publish time.
- [x] **CC-6** Audience naming — **Light.** One paragraph naming "platform and security teams approving agentic deployments" on the homepage value-prop section, no repetition on SDK / proxy pages.
- [x] **CC-7** mcp-proxy tone — **Align with ecosystem reference register.** Downstream of CC-1; falls out naturally once `block` / `pause` are framed as proxy features.
[...]

PR #309, opened 2026-05-02
docs(sdk-py): fix create_receipt and sign_receipt calling conventions

ojongerius:
## Summary

The Python examples on `getting-started/quick-start.mdx` and `sdk-py/overview.mdx` called `create_receipt` with loose kwargs and `sign_receipt` with two positional arguments. Both raise `TypeError` against the shipped API:

- `create_receipt` takes a single `CreateReceiptInput` argument, not loose kwargs (`sdk/py/src/agent_receipts/receipt/create.py:59`).
- `sign_receipt` requires `verification_method` as a third positional argument (`sdk/py/src/agent_receipts/receipt/signing.py:69-73`).
- `sdk-py/overview.mdx` example was also missing the required `chain` field.

Both corrected examples were executed against the real SDK to confirm they run end-to-end.

## Test plan

- [x] `pnpm build` in `site/` passes locally
- [x] Both corrected examples executed against `agent-receipts` via `uv run python -c`, producing valid signed receipts
- [ ] CI green

Refs site action plan #291 (B3, B4, B6, B7).

### DIFF
--- a/site/src/content/docs/getting-started/quick-start.mdx
+++ b/site/src/content/docs/getting-started/quick-start.mdx
@@ -84,31 +84,34 @@ pip install agent-receipts
 
 ```python
 from agent_receipts import (
+    CreateReceiptInput,
     create_receipt,
     generate_key_pair,
-    sign_receipt,
     hash_receipt,
+    sign_receipt,
 )
 
 keys = generate_key_pair()
 
 unsigned = create_receipt(
-    issuer={"id": "did:agent:my-agent"},
-    principal={"id": "did:user:alice"},
-    action={
-        "type": "filesystem.file.read",
-        "risk_level": "low",
-        "target": {"system": "local", "resource": "/docs/report.md"},
-    },
-    outcome={"status": "success"},
-    chain={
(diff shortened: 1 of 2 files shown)


ojongerius:

Both inline comments applied in `10a074e`:

### Comment 1 — `receipt` variable holds the unsigned receipt

**Applied.** Renamed `receipt` → `unsigned` (and updated the `sign_receipt(unsigned, ...)` call) so the variable name reflects what `create_receipt()` actually returns. Now matches the pattern used in `getting-started/quick-start.mdx`.

### Comment 2 — `private_key` undefined in the snippet

**Applied.** Added `generate_key_pair` to the imports, called `keys = generate_key_pair()`, and changed `private_key` → `keys.private_key` in the `sign_receipt` call. The snippet now runs end-to-end as the SDK landing-page example should.

Verified by executing the revised snippet against the real SDK via `uv run python -c` — produces a valid signed receipt with the canonical `u`-prefixed `proofValue`. Copilot's follow-up review found no new issues.

PR #313, opened 2026-05-03
docs: roll out ADR-0012 parameterDisclosure rename across site

ojongerius:
## Summary

[ADR-0012](https://github.com/agent-receipts/ar/blob/main/docs/adr/0012-payload-disclosure-policy.md) renamed the config key `parameterPreview` → `parameterDisclosure` and the receipt field `parameters_preview` → `parameters_disclosure`. The renames already shipped in `sdk/py` and `openclaw/installation.mdx` but the rest of the site still referenced the deprecated names.

ADR-0012 also commits to **operator-controlled** (not user-controlled) disclosure config — independent of the not-yet-shipped encrypted envelope architecture. Two prose statements that implied user control were corrected to match the canonical position.

### Changes

| File | Change |
|---|---|
| `openclaw/cli-reference.mdx` | jq filter field name; note prose; broken anchor `/openclaw/installation/#parameter-preview` → `/openclaw/installation/#parameter-disclosure` (matches existing `## Parameter disclosure` heading on the installation page) |
| `blog/openclaw-plugin-deep-dive.mdx` | Rename `parameterPreview` → `parameterDisclosure` and `parameters_preview` → `parameters_disclosure` throughout (heading, prose, JSON example) |
| `specification/overview.mdx` | `"human principal controls what is disclosed"` / `"user-controlled previews"` → `"operator controls"` / `"operator-controlled disclosures"` |
| `index.mdx` | Same prose framing update on the homepage |

### Out of scope

The blog post still describes the plaintext-in-body disclosure model (where `parameters_disclosure` is a `dict[str, str]`). ADR-0012 supersedes that with an encrypted envelope, but envelope implementation is still **Status: Proposed (Phase A)** — the field is currently a plaintext map in shipped code. Rewriting the blog to describe ciphertext today would misrepresent shipped behavior. Track 3 should revisit once the envelope architecture ships.

The intentional deprecation note in `openclaw/installation.mdx:122-123` retains both old and new names — that's the canonical migration callout and stays as-is.

## Test plan

- [x] `pnpm build` in `site/` passes locally
- [x] `grep -rn "parameterPreview\|parameters_preview\|#parameter-preview" site/` returns only the deprecation note in `openclaw/installation.mdx`
- [x] Anchor target verified: `openclaw/installation.mdx` line 81 has `## Parameter disclosure` which Starlight slugifies to `parameter-disclosure`
- [ ] CI green

Refs site action plan #291 (B14, CC-8, CC-11).

### DIFF
--- a/site/src/content/docs/blog/openclaw-plugin-deep-dive.mdx
+++ b/site/src/content/docs/blog/openclaw-plugin-deep-dive.mdx
@@ -153,11 +153,11 @@ The `receipts` command produces a human-readable summary table for auditing.
 
 ---
 
-## parameterPreview: operator-controlled privacy
+## parameterDisclosure: operator-controlled privacy
 
 By default, the plugin stores only the parameters hash. You cannot reconstruct what arguments were passed to `exec`. This is deliberate — agents frequently receive secrets, tokens, or sensitive data in their tool call parameters.
 
-If you need more forensic detail, set `parameterPreview` in `openclaw.json`:
+If you need more forensic detail, set `parameterDisclosure` in `openclaw.json`:
 
 ```jsonc
 {
@@ -165,7 +165,7 @@ If you need more forensic detail, set `parameterPreview` in `openclaw.json`:
     "entries": {
       "openclaw-agent-receipts": {
         "config": {
-          "parameterPreview": "high"
+          "parameterDisclosure": "high"
         }
       }
     }
@@ -173,7 +173,7 @@ If you need more forensic detail, set `parameterPreview` in `openclaw.json`:
 }
 ```
 
-With `parameterPreview: "high"`, high-risk and critical actions include a `parameters_preview` field alongside the hash:
+With `parameterDisclosure: "high"`, high-risk and critical actions include a `parameters_disclosure` field alongside the hash:
 
 ```json
 {
@@ -181,7 +181,7 @@ With `parameterPreview: "high"`, high-risk and critical actions include a `param
     "type": "system.command.execute",
     "risk_level": "high",
     "parameters_hash": "sha256:9c84a8c9e89a07ff323b0ad52972f148b7f2f5240817f2d9f9892ca514b4522c",
-    "parameters_preview": {
+    "parameters_disclosure": {
       "command": "echo \"Testing agent-receipts plugin fix\""
     }
   }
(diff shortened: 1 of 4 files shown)


ojongerius:

Thanks for the review. Triage and disposition for each of the 5 inline comments:

### Comment 1 — `parameterDisclosure: "high"` (line 168) vs `parameterDisclosure: true` (line 226)

**Keep as-is.** The two examples serve different rhetorical purposes intentionally:

- Line 168 demonstrates the recommended-default `"high"` mode; the table at lines 193–198 enumerates the full value space (`false` / `true` / `"high"` / array).
- Line 226 references `true` in the *"What's absent vs. MCP Proxy"* comparison paragraph, where the worst-case "all actions disclosed" mode is the most pointed framing for the proxy comparison ("even with the most-disclosing setting, the plugin still stores less than the proxy").

Switching them to a single value would either hide the multi-mode design or weaken the comparison.

### Comment 2 — "`parameters_disclosure` is empty by default" wording

**Applied** in `ca01534`: *"empty by default"* → *"omitted by default"*, and *"is only populated when"* → *"is only included when"*. Verified against `sdk/py/src/agent_receipts/receipt/create.py:79-80` — the field is only added to `action_data` when not `None`, so it's omitted from canonical JSON, not present-but-empty. Good catch.

### Comment 3 — jq example targets `parameters_disclosure`

**Partially applied** in `4b3310f`, plus follow-up issue #315.

You correctly flagged that the filter targets `parameters_disclosure`. The deeper bug is that the openclaw CLI's `receipts --json` projection flattens each receipt to `{id, action, risk, target, status, sequence, chain_id, timestamp}` — `parameters_disclosure` isn't included at any depth, so even your suggested `.action.parameters_disclosure` path wouldn't match. The example has been broken since before this rename PR.

In this PR I rewrote the example to filter on `.risk` (a field the CLI actually projects) so the page now shows a working filter, and replaced the note with one that explains the projection limitation and points readers to `inspect <id>` for full receipt bodies. The deeper "extend the CLI projection" fix is tracked in #315 — that change belongs in the openclaw repo, not this docs sweep.

### Comments 4 & 5 — *"Sensitive data never appears in receipts"*

**Rejected; keeping "never".** This is a load-bearing privacy promise backed by:

- The schema description in `spec/schema/agent-receipt.schema.json`: *"Operators MUST NOT include secrets, PII, or other sensitive payload data"*.
- `AGENTS.md` security policy: *"Never store plaintext secrets in receipts"*.
- The SDK type's `Field(description=...)` MUST-NOT language.

`parameters_disclosure` is for non-secret operational context (allowlisted by operator config). The "never" framing matches the protocol's normative position, the same way an audit system's "PII never appears" claim is normative even though misuse is theoretically possible. Softening to "should not" would muddy the privacy story for no real gain.

If a positioning shift is wanted later, that's a Track 3 / pitch.md decision, not a mechanical doc fix.

---

Two real bugs surfaced during this triage that landed as separate issues:

- #315 — openclaw `--json` projection doesn't include `parameters_disclosure`
- #316 — sdk-py `_version.py` drift (caught on PR #314)

ojongerius:
## Second-round disposition (Copilot review at `4215718321`, 3 comments)

No additional changes needed. State at `4b3310f` is current.

### Comment 1 — jq filter fails when `parameters_disclosure` omitted

**Already addressed in `4b3310f`.** That commit rewrote the jq example from `select(.parameters_disclosure.command | strings | contains("rm"))` to `select(.risk == "high")`, eliminating the `parameters_disclosure` index entirely. The accompanying note now also explains that the openclaw CLI's `--json` projection doesn't include `parameters_disclosure` (tracked in #315). The original concern is moot.

### Comments 2 & 3 — *"Sensitive data never appears in receipts"* on `specification/overview.mdx` and `index.mdx`

**Re-issue of first-round Comments 4 & 5; rejected.** Same content already triaged in the first-round disposition above. The phrase is a normative privacy promise backed by:

- The schema description (`spec/schema/agent-receipt.schema.json`): *"Operators MUST NOT include secrets, PII, or other sensitive payload data"*.
- `AGENTS.md` security policy: *"Never store plaintext secrets in receipts"*.
- The Python SDK type's `Field(description=...)` MUST-NOT language.

Operator disclosures (`parameters_disclosure`) are explicitly **not** sensitive data per the schema's contract — operators are forbidden from putting secrets/PII there. Softening "never" to "should not" weakens a load-bearing privacy promise without providing accuracy gain, since the protocol-level commitment is "never" by design.

If the positioning shifts later (e.g., we acknowledge that misuse is theoretically possible), that's a Track 3 / pitch.md decision, not a mechanical doc fix.

Issue #236, opened 2026-04-25
Implement ADR-0010: Daemon process separation for signing and storage


ojongerius:
## Summary

Implement the daemon process separation described in [ADR-0010](https://github.com/agent-receipts/ar/blob/main/docs/adr/0010-daemon-process-separation.md). The current in-process architecture allows agents to forge, suppress, or tamper with their own receipts. This work moves signing and storage into a separate `agent-receipts` daemon running as its own OS user, and reduces every emitter (OpenClaw plugin, MCP proxy, SDK) to a thin fire-and-forget IPC client.

## Background

See ADR-0010 for the full rationale. The short version: an agent auditing itself is not a meaningful audit. The daemon separation restores the tamper-evidence property and collapses N independent crypto/storage stacks into one shared chain.

### 1. IPC transport layer
- [ ] Unix domain socket (`SOCK_SEQPACKET`) at `/run/agentreceipts/events.sock` (Linux) and `/var/run/agentreceipts/events.sock` (macOS)
- [ ] Named pipe (`\\.\pipe\agentreceipts-events`) on Windows with equivalent ACL semantics
- [ ] Socket/pipe path configurable via env var (e.g. `$XDG_RUNTIME_DIR/agentreceipts/events.sock` for unprivileged installs)
- [ ] Non-blocking send on emitter side; `EAGAIN` increments a local drop counter instead of blocking

### 2. Daemon process (`agent-receipts-daemon`)
- [ ] Sole owner of Ed25519 signing keys and SQLite database
- [ ] Peer credential capture at connection-accept time:
  - Linux: `SO_PEERCRED` (`uid`, `gid`, `pid`)
  - macOS: `LOCAL_PEERCRED` + `LOCAL_PEEREPID`
  - Windows: `GetNamedPipeClientProcessId` + `OpenProcessToken` for SID and integrity level
- [ ] Executable path resolution: `/proc/<pid>/exe` (Linux), `proc_pidpath` (macOS), `QueryFullProcessImageName` (Windows)
- [ ] RFC 8785 canonicalization (moved exclusively from emitters to daemon)
- [ ] Hash-chaining and Ed25519 signing (seq, prev_hash, ts_recv, peer, id added by daemon)
- [ ] SQLite persistence with `events_dropped` synthetic receipt when a gap is recorded
- [ ] DB permissions: `0640` owner `agentreceipts`, group `agentreceipts-read`; public key `0644`

### 3. Thin emitter refactor
[...]

Issue #155, opened 2026-04-13
Document threat model and trust boundaries


ojongerius:
## Problem

Security-conscious users evaluating the proxy need to understand what it does and doesn't protect against. Currently there's no explicit threat model documentation, which means:

- **Security leads** can't assess the proxy for their environment without reverse-engineering the trust assumptions
- **White-hat reviewers** discover trust boundaries by probing rather than reading, which wastes their time and may lead to "vulnerabilities" that are actually known limitations
- **Compliance teams** can't reference a threat model when writing audit narratives

The proxy makes specific trust assumptions (e.g. the host isn't compromised, the signing key is protected) that should be stated upfront, not discovered.

## Proposal

Create a `THREAT_MODEL.md` document covering:

### Trust boundaries
- The proxy trusts the host OS and filesystem
- The proxy trusts the MCP client (Claude Desktop/Code) to faithfully relay messages
- The proxy does NOT trust the upstream MCP server (that's the point — it's witnessing the server's behaviour)

### In-scope threats (what receipts protect against)
- Tampered receipts (signature verification catches modifications)
- Missing receipts (chain gap detection catches deletions)
- Replayed receipts (sequence numbers and timestamps detect replay)
- Reordered receipts (hash chain catches reordering)
- Sensitive data in storage (redaction patterns remove known secrets)

### Out-of-scope threats (explicitly not protected)
- Compromised host — attacker with filesystem access can steal the signing key and forge receipts
- Replaced proxy binary — a malicious binary can do anything (see #151 for mitigation)
- Colluding MCP client — if the client doesn't send calls through the proxy, there's no receipt
- Network-level attacks on localhost — not applicable for stdin/stdout transport

### Mitigations and roadmap
- Key management improvements (HSM support, key rotation)
- Remote attestation (future)
- DID-based identity resolution (ADR-0007, #46)

## Acceptance criteria

- [ ] `THREAT_MODEL.md` exists in the repo root or `docs/` directory
- [ ] Document covers: trust boundaries, in-scope threats, out-of-scope threats, and mitigations/roadmap
- [ ] Each threat has a brief description and the mechanism that addresses it (or "out of scope" with rationale)
- [ ] Document is linked from the main README
- [ ] Document references related issues (#151 supply chain, #156 key permissions, #46 DIDs)
- [ ] Reviewed by at least one person with security experience before merge
- [ ] Docs site updated with a "Security" section linking to the threat model

jagmarques:
[...]

Issue #32, opened 2026-04-04
Design algorithm agility for post-quantum migration


ojongerius:
## Problem

The protocol currently hardcodes Ed25519 for signing and SHA-256 for hashing. Both are solid choices today, but:

- **NIST recommends migrating from ECC to ML-DSA by 2030**, with quantum-ready cryptography mandatory for government agencies after 2035
- The industry is moving toward **hybrid signatures** (ML-DSA + Ed25519) as a transition strategy
- ML-DSA signatures are ~5x larger than Ed25519 (~2.4kB vs ~64 bytes), which has implications for receipt storage and transport
- SHA-256 is secure and universal, but SHA-3 and BLAKE3 offer improved properties — the protocol should not assume SHA-256 forever

Hardcoding algorithms now without an upgrade path means a breaking protocol change later.

## Proposal

Design algorithm agility into the spec so the protocol can migrate without breaking existing receipts:

1. **Algorithm identifiers in receipts** — ensure the signing algorithm and hash algorithm are explicitly declared in each receipt (not just implied), so verifiers know which algorithm to use
2. **Algorithm negotiation/versioning** — define how new algorithms are introduced alongside existing ones (e.g., protocol version bumps, algorithm suites)
3. **Hybrid signature support** — plan for composite signatures (ML-DSA + Ed25519) as the likely transition step
4. **Receipt format impact analysis** — ML-DSA signatures are ~40x larger than Ed25519; assess impact on storage, transport, and chain linking
5. **Migration timeline** — align with NIST guidance (ML-DSA by 2030) and define when the protocol should support post-quantum algorithms

## Non-goals

- Actually implementing ML-DSA or other post-quantum algorithms now — this is about designing the spec to support them later
- Weakening the current Ed25519/SHA-256 requirements — these remain the only supported algorithms until the spec explicitly adds alternatives

## References

- [NIST PQC Standardization (FIPS 204 — ML-DSA)](https://postquantum.com/post-quantum/cryptography-pqc-nist/)
- [Composite ML-DSA + Ed25519 signatures (IETF draft)](https://datatracker.ietf.org/doc/html/draft-ietf-lamps-pq-composite-sigs-15)
- [Post-Quantum Cryptography for Verifiable Credentials (TechRxiv)](https://www.techrxiv.org/users/973090/articles/1346363/master/file/data/main/main.pdf)
[...]

PR #319, opened 2026-05-03
docs(adr): add ADR-0015 (key rotation, BYOK, external anchoring) — Proposed

### LINKED ISSUE #307: ADR-0015: Key rotation, BYOK abstraction, and external anchoring (Proposed)
## Problem

The daemon (ADR-0010, #236) holds a single Ed25519 signing key. We currently have no specified story for:

1. **Rotating that key** without breaking historical chain verification
2. **Where the key lives** — the daemon assumes a file on disk with `0600` perms, which is fine for the local-developer case but is not the answer enterprise operators will accept (HSM, cloud KMS, customer-managed key store)
3. **What happens when the daemon itself is compromised** — chain verification catches mid-chain tampering but not tail truncation (#171) or a forged rotation history. Without external anchoring of rotation events and periodic chain commitments, "the daemon protects the chain" is a circular argument

This ADR scopes the design that closes those three gaps. Implementation is split into follow-on issues; this issue is the Proposed-status ADR document only.

## Proposal

Two primitives, designed together because their interfaces interlock.

### 1. BYOK abstraction (`KeySource` interface)

The daemon reads its signing key from a configurable backend rather than a fixed filesystem path. Backends:

- **File** (default) — current behaviour, `~/.agent-receipts/signing.key`
- **PKCS#11** — HSM / smartcard / TPM via standard interface
- **Cloud KMS** — adapter pattern: AWS KMS, GCP KMS, Azure Key Vault. Key never leaves the KMS; daemon submits canonical bytes for signing

The interface is the deliverable in this ADR. Adapters land in follow-on issues. Algorithm-agnostic by design (see #32 for post-quantum).

### 2. External rotation anchoring

When the daemon rotates its signing key, the rotation event is written to:

- The local chain (always — `key_rotated` synthetic receipt with old/new public key fingerprints)
- An optional external sink configured by the operator (webhook, S3 with object-lock, transparency log, customer SIEM)

The external sink is what makes post-compromise integrity claims defensible. Without it, an attacker who reaches the daemon's signing key can also forge a plausible rotation history alongside forged receipts. With it, the rotation timeline lives somewhere the daemon does not control, and forged tails are detectable by checkpoint mismatch.

This is the structural answer to "what if the daemon itself is compromised" and it is the property centralized credential-vault tools cannot match by their architecture.

### Periodic chain commitments (related, scope decision required)

The same external-sink mechanism naturally extends to periodic chain-tip commitments — writing the current `seq` and `prev_hash` to the external sink at intervals. This addresses tail truncation (#171) directly. **Open question for the ADR:** scope this here as part of "external anchoring", or split into a sibling ADR. Lean toward in-ADR — same sink, same operator config surface, same threat lineage.

## Acceptance criteria

This issue is the **ADR document only**. Implementation is split out.

- [ ] `docs/adr/0015-key-rotation-byok-anchoring.md` exists, status: Proposed, Nygard format
- [ ] Document specifies the `KeySource` interface contract (operations, error semantics, init/teardown) without implementing adapters
- [ ] Document specifies the rotation event schema (canonical bytes, what's signed by old vs new key, how verifiers chain through rotations)
- [ ] Document specifies the external-anchor write contract (event types: rotation, checkpoint; transport-agnostic; failure semantics — does daemon block, queue, or drop on sink unavailability?)
- [ ] Decides scope question on periodic chain commitments (in-ADR or sibling ADR)
- [ ] Cross-references #155 (threat model — the "why"), #171 (tail truncation — concrete threat addressed), #236 (ADR-0010 daemon — the substrate), #32 (algorithm agility — interface constraint)
- [ ] Updates `docs/adr/README.md` index

## Numbering note

Originally drafted as "ADR-0008" but `0008` is taken by `0008-response-hashing-and-chain-completeness.md`. Renumbered to ADR-0015 (next free after 0014 codex-hook-channel).

## Out of scope (follow-on issues, not blocking this ADR)

- PKCS#11 adapter implementation
- AWS KMS / GCP KMS / Azure Key Vault adapters
- Specific external-sink integrations (S3 object-lock, sigstore, Splunk, etc.)
- CLI for triggering manual rotation
- Migration path for existing v1 keys
- BYOK policy enforcement (which key sources are permitted in which deployment tiers)

## Related

- ADR-0010 / #236 — daemon process separation (substrate this ADR sits on)
- #155 — threat model (motivation; will be refreshed to reference this ADR)
- #171 — tail truncation in chain verification (specific threat external anchoring addresses)
- #32 — algorithm agility (`KeySource` interface must not preclude PQ migration)
- ADR-0001 — Ed25519 signing (key now sourced via `KeySource` rather than fixed path)

ojongerius:
## Summary

Translates issue #307 into a Nygard-format Proposed ADR. Closes the three gaps left by ADR-0010 (#236):

1. **Key rotation** without breaking historical chain verification
2. **Where the key lives** — file / PKCS#11 / cloud KMS, via a `KeySource` interface
3. **Post-compromise integrity** — an external sink the daemon does not control

Two primitives, designed together because their interfaces interlock:

### `KeySource` interface (BYOK abstraction)

Operations: `Sign` / `PublicKey` / `Rotate` / `Init` / `Teardown`. Algorithm-agnostic so [#32](https://github.com/agent-receipts/ar/issues/32) (algorithm agility) is reachable without a redesign. Backends in scope: file (current default), PKCS#11, cloud KMS. Adapters land in follow-on issues.

### External anchor write contract

Rotation events and periodic checkpoints written to an operator-configured sink. Three failure modes (`block` / `queue` / `drop`) exposed as operator config, not architecture. The conditional integrity claim — *"chain history survives daemon-key compromise IF rotation events are anchored externally"* — is now load-bearing and explicitly stated.

### Scope decision on periodic chain commitments

**In-ADR.** Same sink, same config surface, same threat lineage as rotation anchoring. Phase A ships rotation anchoring; Phase B adds checkpoint anchoring with the threat model ([#155](https://github.com/agent-receipts/ar/issues/155)) updating concurrently to assert the conditional integrity guarantee.

## Numbering note

Originally drafted as "ADR-0008" in issue #307 but `0008` is already taken by `0008-response-hashing-and-chain-completeness.md`. Renumbered to **ADR-0015** (next free after `0014-codex-hook-channel.md`). The issue title and body have been updated to match.

## Cross-references

- [ADR-0001](https://github.com/agent-receipts/ar/blob/main/docs/adr/0001-ed25519-for-receipt-signing.md) — current default algorithm; narrowed from "baked in" to "current default"
- [ADR-0007](https://github.com/agent-receipts/ar/blob/main/docs/adr/0007-did-method-strategy.md) — public-key resolution path for fingerprints
- [ADR-0010](https://github.com/agent-receipts/ar/blob/main/docs/adr/0010-daemon-process-separation.md) / #236 — substrate this ADR sits on
- [ADR-0012](https://github.com/agent-receipts/ar/blob/main/docs/adr/0012-payload-disclosure-policy.md) — separate forensic keypair, separate lifecycle (informs the "do not reuse keys across purposes" property)
- #155 — threat model (motivation; will be refreshed to reference this ADR)
- #171 — tail truncation in chain verification (specific threat external anchoring addresses)
- #32 — algorithm agility (`KeySource` interface must not preclude PQ migration)

## Test plan

- [x] `docs/adr/0015-key-rotation-byok-anchoring.md` exists, status: Proposed, Nygard format
- [x] Document specifies `KeySource` interface contract (operations, error semantics, init/teardown) without implementing adapters
- [x] Document specifies rotation event schema (canonical bytes, what's signed by old vs new key, how verifiers chain through rotations)
- [x] Document specifies external-anchor write contract (event types: rotation, checkpoint; transport-agnostic; failure semantics: block/queue/drop)
- [x] Decides scope question on periodic chain commitments (**in-ADR**, with phasing)
- [x] Cross-references #155, #171, #236, #32, ADR-0001, ADR-0007, ADR-0012
- [x] Updates `docs/adr/README.md` index
- [ ] CI green

Closes #307.


ojongerius:

Thanks for the review. All 8 comments triaged; 5 applied in `7793222`, 3 substantive design changes deferred for discussion before applying.

### Applied in `7793222`

#### Comment 2 — `queue` mode integrity framing
Tightened the failure-modes table cell and added a new Negative-consequences bullet. The prior phrase *"tolerates transient sink outages without losing integrity guarantees once the sink recovers"* was too strong — events in the outbox between local-write and sink-acknowledge are not yet anchored, and a daemon compromised mid-window can rewrite the outbox. Now states explicitly: events in the outbox are tracked locally but not yet anchored until acknowledged, with `block` as the alternative for tightly compliance-bound deployments.

#### Comment 4 — encoding cross-reference
Verified: ADR-0001 (line 51) defines the `u`-prefix base64url encoding choice; ADR-0003 defers to it; ADR-0012 does not define it. Citation corrected from ADR-0012 → ADR-0001.

#### Comment 5 — `KeySource` return shapes inconsistent
Real ambiguity: bullet list said `Sign → signature, error` but the prose said Sign returns "opaque signature bytes plus an algorithm tag." Aligned both on explicit tuples — `Sign → (signature bytes, algorithm tag, error)` and `Rotate → (old fingerprint, new fingerprint, algorithm tag, transcript bytes, error)`. Added a sentence in the `Rotate` description handling cross-algorithm rotation explicitly (relates to Comment 7 below).

#### Comment 6 — fingerprint canonical form underspecified
Genuine interop hazard for a Proposed-status design contract. Added a *"Fingerprint canonical form"* paragraph: SHA-256 over raw public key bytes per algorithm (Ed25519: 32-byte public key per RFC 8032 §5.1.5). SPKI/PEM wrappers and backend handles (KMS key IDs, PKCS#11 object handles) MUST NOT be hashed — those would produce different fingerprints for the same underlying key across adapters.

#### Comment 8 — verification flow conflated fingerprints with public keys
Reframed (PARTIAL — your literal fix would have over-promised DID coverage). Rewrote the numbered steps to resolve `old_key_fingerprint` via the configured key registry (DID per ADR-0007, file, or KMS reference) *first*, then verify the rotation event's signature with the resolved public key. Added a note that rotation events SHOULD carry a `proof.verificationMethod` mirroring the receipt envelope; the fingerprint fields exist for fast chain-traversal indexing and for offline verification when DID resolution is unavailable.

### Deferred for discussion (open follow-up commits)

#### Comment 1 — checkpoint payload `prev_hash` vs `tip_hash`
You correctly flag a semantic distinction: committing to the predecessor's link vs. the tip's own canonical hash. Functional impact for tail-truncation detection is subtler than your write-up suggests — the older legitimate checkpoint already in an append-only sink catches truncation via `seq` regression regardless of which hash is anchored. But "commits to the tip" is the cleaner statement, and `tip_hash` is more semantically faithful to what the anchor is asserting. Holding on this one for explicit review of the protocol semantics before changing.

#### Comment 3 — retain old public keys, not private keys
You're right that the bracketed parenthetical comparing this to ADR-0012's "old forensic private keys retained forever" is misleading: forensic *decryption* legitimately needs the old private key to read historical encrypted payloads, but signing-rotation verifiers only need the old *public* key. Telling operators to retain old private signing keys forever extends the secret blast radius needlessly. Holding to confirm the resulting wording (especially the destruction guidance — "SHOULD destroy retired private signing keys once rotation is anchored" is the natural phrasing).

#### Comment 7 — split `algorithm` into `old_algorithm` + `new_algorithm`
For same-algorithm rotation the two values are equal and the single field reads naturally; for cross-algorithm rotation (Ed25519 → ML-DSA per #32) the verifier needs both to know which algorithm verifies the rotation event itself versus subsequent receipts. Holding to confirm the schema change before applying.

---

Comments 4–8 fixes verified against the ADR text and related ADRs (0001, 0007). Comments 2 was a content-substantive integrity-framing fix and is in.

ojongerius:
## Follow-up: Comments 1, 3, 7 applied in `7b908a6`

The three deferred design changes are now in.

### Comment 1 — checkpoint commits to tip, not predecessor

External-anchor checkpoint payload changed from `(seq, prev_hash, public_key_fingerprint)` to `(seq, tip_hash, public_key_fingerprint)`. `tip_hash` is the SHA-256 of the canonical bytes of the most recently appended receipt itself (per ADR-0001's `u`-prefix base64url encoding) — not its `prev_hash`. Verifiers now detect tail truncation as a mismatch on either `seq` or `tip_hash`. Functional impact for detection is small — older legitimate checkpoints still catch truncation via seq regression — but the wording now matches what the anchor is actually asserting. The threat model (PR #320) was updated in `3384748` to keep the `(seq, tip_hash, public_key_fingerprint)` reference consistent.

### Comment 3 — retain old public keys; destroy retired private keys

The Negative-consequences bullet was rewritten. Previous text told operators to retain old private keys forever and compared this to ADR-0012's forensic keypair. That comparison was wrong: ADR-0012's *decryption* key legitimately needs private-key retention to read historical encrypted payloads, but signing verifiers only ever need the public key. Now: retired private signing keys SHOULD be destroyed once rotation is anchored externally — reducing private-key blast radius is a positive consequence of rotation, not a tax on it.

### Comment 7 — split `algorithm` into `old_algorithm` + `new_algorithm`

Schema table now has two fields:

- `old_algorithm` — algorithm tag of the outgoing key. Used to verify the rotation event's own signature (since `signed_with: "old"`).
- `new_algorithm` — algorithm tag of the incoming key. Used to verify subsequent receipts. Equal to `old_algorithm` in same-algorithm rotations; differs only across cross-algorithm migrations (Ed25519 → ML-DSA per #32).

Verification flow steps 1 and 3 updated to reference the two fields explicitly.

---

All 8 Copilot comments addressed: 5 in `7793222`, 3 in `7b908a6`. Threat-model PR #320 has the matching consistency change in `3384748`.

ojongerius:
## Second-round disposition (Copilot review at `4215862123`, 11 comments)

5 of 11 applied in `d80f411`. The other 6 are duplicates of the first-round comments already addressed in `7793222` + `7b908a6`.

### Applied in `d80f411`

1. **File backend path** uses `~/.agent-receipts/signing.pem` (and `signing.pem.pub`), not `signing.key` — matches the existing mcp-proxy `init` convention.
2. **Checkpoint payload** now scopes by `issuer.id`: `(issuer.id, seq, tip_hash, public_key_fingerprint)`. Without this, checkpoints from multiple daemons (multi-host, test environments) cannot be distinguished in a shared sink.
3. **`payload bytes` canonicalization** specified — must be RFC 8785 canonical JSON (per ADR-0002 / ADR-0005 / ADR-0009). Same rule receipts use; without it, sink-adapter bugs could produce non-verifiable anchor payloads.
4. **Sink-as-anchor properties** named explicitly: append-only retention and sink-controlled ordering/timestamps. A bare webhook POST without these is *transport*, not an anchor. Operators who want the post-compromise integrity claim must choose a sink that delivers both.
5. **`block` mode wording corrected.** Per ADR-0010, emitters are non-blocking and drop events via the `EAGAIN` mechanism — "block" doesn't actually block the emitter; the daemon refuses to accept events, and emitters drop them, recorded as `events_dropped` synthetic receipts. The mode's value is *no silent loss* (gaps in-chain), not zero loss. The original wording overstated availability cost as a wash for "maximum integrity."

### Duplicates of first-round comments (already addressed)

The remaining 6 second-round comments restate first-round Comments 1, 2, 3, 4, 6, 7, 8 — all addressed in `7793222` (`u`-prefix encoding cross-ref → ADR-0001; KeySource return tuples; fingerprint canonical form; verification flow registry resolution) and `7b908a6` (`tip_hash`; retain old public not private keys; split `algorithm` into `old_algorithm` + `new_algorithm`).

The second review's `commit_id` field reports `7404ebf` (the original PR HEAD), which is misleading — it appears Copilot's bot was triggered before noticing my fix commits. The comments themselves describe the same defects flagged in round one. No additional action needed for those.

---

Apply 5 of 11. Already addressed 6 (duplicates of round 1). Reject 0. Latest state on the PR is `d80f411`.

PR #320, opened 2026-05-03
docs(threat-model): draft v2 daemon-aware threat model

ojongerius:
## Summary

Translates the scope-update [comment on #155](https://github.com/agent-receipts/ar/issues/155#issuecomment-4364761134) into a `docs/threat-model.md` document. Forward-looking framing: the body describes the v2 architecture introduced by [ADR-0010](https://github.com/agent-receipts/ar/blob/main/docs/adr/0010-daemon-process-separation.md) (daemon process separation) and ADR-0015 (key rotation, BYOK, external anchoring — landing in #319), with a **v1-caveats appendix** structured to be deleted atomically when ADR-0010 implementation ships (#236).

## External reviewer feedback folded in

Two items from [@jagmarques (comment 4341408244)](https://github.com/agent-receipts/ar/issues/155#issuecomment-4341408244) that the original scope update missed:

1. **Honest system clock** — listed in *"Trust still rests on"*. Replay detection via timestamps and the temporal ordering of rotation events both assume the host clock; an attacker who controls the clock can backdate events.
2. **Lying MCP client / TOCTOU** — distinct out-of-scope item, separate from the existing *"colluding client that bypasses the proxy"* case. A compromised client that *does* relay through the proxy but constructs different payloads for the model versus the proxy is outside the trust boundary; the receipt witnesses what the client sent through the proxy, not what the model actually saw.

## What we explicitly claim

The document names the conditional integrity guarantee load-bearingly:

> Receipts are cryptographically tamper-evident… The chain's historical integrity survives daemon-key compromise **if and only if rotation events and periodic checkpoints are anchored to a sink the daemon does not control.**

This conditional is the criterion sheet for "is the daemon done?" — it's what ADR-0010 implementation must satisfy, and what ADR-0015 specifies the mechanism for.

## Structure

1. Trust boundaries (restructured around the daemon)
2. What we explicitly claim (the conditional, in plain English)
3. In-scope threats (additions: compromised agent process — now defended; tail truncation — Phase B mitigation; forged rotation history — Phase A mitigation)
4. Out-of-scope threats (refinements: bounded daemon compromise, not total; new lying-MCP-client item)
5. **v1 caveats (current state)** — atomic block that deletes when #236 ships
6. Mitigations and roadmap (table linking to ADR-0007, ADR-0010, ADR-0015, plus #32, #151, #156)

## Sequencing

[...]

PR #321, opened 2026-05-03
docs(adr): add ADR-0016 (mcp-proxy audit encryption at rest) — Proposed

### LINKED ISSUE #61: ADR for cryptographic design choices in mcp-proxy
## Summary

Write an ADR documenting the encryption design in mcp-proxy:

- Algorithm: AES-256-GCM
- Key derivation: Argon2id (time=1, mem=64MB, threads=4)
- Salt strategy: per-installation random salt persisted in SQLite metadata table
- Assumptions and limitations (e.g., single-passphrase model, no key rotation yet)

### Why

The original implementation used a hardcoded salt — a reasonable trade-off at the time, but underdocumented. When the threat model changed (audit data moving to remote storage), the design gap was only caught during code scanning review. An ADR would have made the trade-off explicit and flagged it as provisional.

ADR infrastructure already exists in this repo (see `docs/adr/`).

ojongerius:
## Summary

Closes #61. Documents the encryption-at-rest design that already ships in `mcp-proxy/internal/audit/encrypt.go`, in Nygard format, status **Proposed**. The implementation evolved through three salt strategies under a changing threat model and the rationale was scattered across commit messages and a code-scanning finding — this ADR makes the current state readable as a deliberate design rather than an accreted one.

### Design

- **Cipher:** AES-256-GCM with a fresh 96-bit random nonce per encryption. Wire format: `enc:` + base64(nonce ‖ ciphertext ‖ tag).
- **KDF:** Argon2id at the OWASP baseline — `time=1, memory=64 MiB, parallelism=4, output=32 bytes`. Derivation runs once at proxy startup.
- **Salt:** 16 random bytes generated on first use and persisted in the SQLite `metadata` table (`encryption_salt`). Per-installation, never per-row, never constant.
- **Passphrase:** read from `BEACON_ENCRYPTION_KEY`. Empty value = encryption disabled (cleartext + post-redaction). Non-empty = mandatory; no silent fallback to plaintext on encrypt error.

### Why this ADR exists

A previous iteration used a constant compile-time salt. With remote storage of audit data (S3 backups, SIEM replication, network filesystems) on the table, that choice meant a single passphrase compromise could be amortised across every installation via offline precomputation. The fix shipped in `cb6911a` (per-installation random salt) but the trade-off was never written down. ADR-0010 / 0015 are about to land work in this neighbourhood (daemon ownership, key rotation, BYOK), and they need the encryption baseline documented to compose against.

### Known limitations explicitly listed

So future ADRs inherit a stated baseline rather than rediscovering constraints:

- No key rotation (passphrase change today = data unrecoverable).
- Passphrase lives in environment (process listings, crash dumps, `/proc/<pid>/environ`).
- AAD is `nil` — ciphertexts not bound to (receipt UUID, field name).
- Metadata (tool name, timestamps, decisions, intent ids) is not encrypted; verifiers need it.
- No passphrase strength enforcement.
- Per-installation, not per-tenant.

### Numbering note

ADR-0015 is in flight in #319 (key rotation, BYOK, external anchoring), so this ADR took **0016**. The index will carry a 0015 gap until #319 merges. Cross-references to ADR-0015 in this document assume that filename (`0015-key-rotation-byok-anchoring.md`) per the open PR.

## Cross-references

- [ADR-0001](https://github.com/agent-receipts/ar/blob/main/docs/adr/0001-ed25519-for-receipt-signing.md) — receipt signatures independent of this layer
- [ADR-0004](https://github.com/agent-receipts/ar/blob/main/docs/adr/0004-sqlite-for-local-receipt-storage.md) — ciphertexts and salt share the same SQLite file
- [ADR-0008](https://github.com/agent-receipts/ar/blob/main/docs/adr/0008-response-hashing-and-chain-completeness.md) — chain integrity preserved with or without encryption
- [ADR-0010](https://github.com/agent-receipts/ar/blob/main/docs/adr/0010-daemon-process-separation.md) — passphrase model can move into the daemon
- [ADR-0012](https://github.com/agent-receipts/ar/blob/main/docs/adr/0012-payload-disclosure-policy.md) — redaction runs before encryption
- [ADR-0015](https://github.com/agent-receipts/ar/pull/319) — encryption-key analogue (rotation/BYOK) is future work that should reuse the same `KeySource` shape
- #61 — motivating issue

## Test plan

- [x] `docs/adr/0016-mcp-proxy-audit-encryption.md` exists, status: Proposed, Nygard sections (Status / Context / Decision / Security Considerations / Known Limitations / Consequences / Related ADRs)
- [x] Cipher, KDF parameters, salt strategy, passphrase ingestion, and fail-closed semantics each documented with rationale
- [x] Known Limitations section lists each gap explicitly so future ADRs (rotation, BYOK, daemon ownership) compose against a stated baseline
- [x] Cross-references #61, ADR-0001, ADR-0004, ADR-0008, ADR-0010, ADR-0012, ADR-0015
- [x] Updates `docs/adr/README.md` index
- [ ] CI green

### DIFF
--- a/docs/adr/0016-mcp-proxy-audit-encryption.md
+++ b/docs/adr/0016-mcp-proxy-audit-encryption.md
@@ -1,0 +1,134 @@
+# ADR-0016: Audit Store Encryption at Rest (mcp-proxy)
+
+## Status
+
+Proposed
+
+## Context
+
+The mcp-proxy SQLite audit store records the JSON-RPC request and response traffic between MCP clients and MCP servers. After redaction (see [ADR-0012](./0012-payload-disclosure-policy.md)), a redacted view of the raw envelope, the tool-call arguments, the response result, and any error are persisted alongside each receipt. Even post-redaction, these fields routinely contain commercially or operationally sensitive content: filenames, branch names, ticket bodies, partial code, search queries, customer-facing strings. Redaction removes credentials and known-sensitive patterns; it is not a substitute for confidentiality of the remainder.
+
+The original threat model assumed the SQLite file lived on the operator's local disk under their primary user account, in which case file-system permissions were sufficient. That assumption no longer holds. Operators ship audit databases off the host: backup to S3, archival to cold storage, replication into a SIEM, mount over network filesystems for centralised review. Once the file moves off the originating machine, every reader of the destination becomes a reader of the audit content, and the local-disk assumption stops paying for itself.
+
+This ADR documents the encryption design that already ships in `mcp-proxy/internal/audit/encrypt.go`. It exists because the implementation evolved through three iterations under a changing threat model, the rationale was scattered across commit messages and a code-scanning finding, and the current state needs to be readable as a deliberate design rather than an accreted one. Three questions had to be answered:
+
+1. **Cipher**: how should an individual field be protected at rest?
+2. **Key derivation**: the input is an operator-supplied passphrase rather than a high-entropy key, so a KDF is required — which one, and tuned how?
+3. **Salt strategy**: hardcoded, passphrase-derived, per-row, or per-installation?
+
+A previous iteration used a constant compile-time salt. That choice was caught during code-scanning review once remote storage entered the picture: with a constant salt, the same passphrase produces the same key across every installation, so an attacker who obtains one passphrase or precomputes a dictionary against the constant salt can attack every other install for free. The salt question is the one that turned the design from "passable for local files" into "needs to hold up under remote storage."
+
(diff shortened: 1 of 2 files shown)


ojongerius:
## Copilot review — disposition (4 comments)

All 4 inline comments are about the same issue: relative links to `./0015-key-rotation-byok-anchoring.md` (or anchors thereof) are dangling on this branch standalone, since ADR-0015 lands via PR #319.

**No doc changes needed in this PR.** Resolution is merge ordering rather than text rewrites:

1. **Merge #319 first.** Once #319 is in `main`, all four relative links resolve cleanly — they're already pointing at the correct future filename (`docs/adr/0015-key-rotation-byok-anchoring.md`).
2. **Then merge #321** (this PR). The ADR index gap from `0014` to `0016` closes the moment #319 lands.

Switching to absolute URLs that point at PR #319 directly would become *wrong* the moment #319 merges (PRs don't render as ADR pages and the URL would be a permanent indirection). Adding a placeholder ADR-0015 file in this PR would be scope creep and a phantom artifact in the index. The right answer is the merge gate.

This mirrors the same pattern used on PR #320 (threat model), which has a relative link to ADR-0015 that resolves once #319 merges. Same merge sequence applies: **#319 → #320 → #321**.

PR #507, opened 2026-05-21
feat(spec-sdks): chain_id binding check in verifier (#477)

### LINKED ISSUE #477: [protocol] Explicit sessionId binding in verifier
## Problem

Nothing prevents replaying a valid receipt from session A into session B. The
signature verifies and the hash chains correctly if constructed carefully.
`sessionId` is present in the credential body covered by the proof, but
verifiers do not currently enforce it explicitly.

## Decision (ADR-0019 § P4)

Verifiers must reject receipts whose `sessionId` does not match the session
being verified. Add explicit sessionId binding check to:

- The verification algorithm in the spec
- All three SDK verifiers

## References

ADR-0019 § P4

ojongerius:
## What

Adds an unconditional `chain.chain_id` binding check to the verification algorithm. All receipts in a verified chain MUST share the same `chain.chain_id` as the first receipt; verifiers reject input that mixes receipts from different chains.

Closes #477 (ADR-0019 § P4 — v1-blocker).

## Why

Before this PR the verifier checked signatures, hash linkage, sequence monotonicity, and the receipt-after-terminal invariant, but it did NOT check that `chain.chain_id` was consistent across the input. An attacker who can forge a valid-looking hash link between two distinct chains could splice them together and pass them off as one logical chain under a single `verifyChain` call.

The new check fires independently of hash linkage, so a forged `previous_receipt_hash` between two chains still causes rejection.

Note on terminology: the issue title says "sessionId", referring to the ADR-0019 § P4 wording. The protocol's actual session/chain identifier is `chain.chain_id` (spec § 4.3.2). The separate optional `action.session_id` field is a different scope and is not what this PR touches. This was confirmed in the issue body.

## Changes

**Spec** (`spec/spec/agent-receipt-spec-v0.1.md`)
- New § 7.3.4 "Chain identifier binding (automatic)" — parallels § 7.3.2 (receipt-after-terminal). Unconditional check, no caller parameter can suppress it. Error must identify the offending index and both `chain_id` values.
- § 4.3.2 `chain.chain_id` row cross-references § 7.3.4.

**TypeScript SDK** (`sdk/ts/src/receipt/chain.ts`)
- New chain_id binding scan, runs before the receipt-after-terminal scan.
- Error: `chain_id mismatch at index N: expected "X", got "Y"`.
- 3 new tests: single-chain baseline, cross-chain splice with forged hash linkage, single off-chain receipt re-signed under a different `chain_id`.

**Go SDK** (`sdk/go/receipt/chain.go`)
- Same check structure, returns the same error shape.
- 3 new tests mirroring TS.

**Python SDK** (`sdk/py/src/agent_receipts/receipt/chain.py`)
- Same check structure.
- 3 new tests mirroring TS.

## Tests

- TypeScript: **279 passed** (3 new)
- Go: full SDK + `cross-sdk-tests` (integration tag) + `mcp-proxy` + `hook` all pass (3 new in `receipt/`)
- Python: **302 passed + 5 skipped** (3 new)
- Cross-SDK integration tests pass (`go test -tags=integration ./...` in `cross-sdk-tests`)

## Pattern

Follows PR #501's pattern (spec + 3 SDKs + verifier + tests, one PR).

## Checklist

- [x] Tests pass for all changed components
- [x] Linter passes (go vet, ruff check + format, biome) — lefthook pre-commit hooks all green
- [x] No real keys or secrets in the diff
- [x] Cross-language tests pass (`cross-sdk-tests` with `-tags=integration`)
- [ ] AGENTS.md updated (N/A — no project structure change)
- [x] Spec changes — § 7.3.4 added per issue #477 / ADR-0019 § P4

## Security

- [x] Touches the verifier surface; the new check is an unconditional rejection rule (parallels § 7.3.2). It does not change any cryptographic primitive, key handling, or signature scheme.
- [x] Inputs validated: each receipt's `chain.chain_id` is compared to that of `receipts[0]`; on mismatch the verifier returns immediately with a structured error. The check runs even when other earlier-detected errors are present so a cross-chain splice cannot hide behind a signature or hash-compute error.
- [x] Edge cases tested: empty chain (no check needed, no traversal), single-receipt chain (no second receipt to compare against), splice at index 2 with valid hash linkage (forged), mismatched middle receipt with valid signature for its own `chain_id`.


ojongerius:
Round 4 follow-ups:

- **Addressed in `c1c4fd6`**: `hashComputeErrorAt` off-by-one (now stored as `i-1` to match message); `loopError` precedence now uses earliest-index logic with sig-vs-hash as tie-break, mirroring Go.
- **Not changing — contradicts round-1 feedback**: chain_id mismatch `brokenAt`. Round 1 explicitly asked to set `brokenAt: i` for this return so it aligns with the error message; round 4 now asks for the opposite (preserve earlier `brokenAt`). The error message clearly identifies the mismatch index, so callers reading both the error and `brokenAt` won't be confused.
- **Deferred to follow-up #513**: chain_id binding tests refactored to isolate the invariant from hash-linkage collateral failure across all three SDKs.

PR #532, opened 2026-05-22
Add hermes-agent plugin for Agent Receipts audit trails

ojongerius:
> ⚠️ **Draft — do not merge.** Pending manual end-to-end verification by the author against a live hermes-agent runtime + agent-receipts daemon. Automated tests (pytest/ruff/pyright) pass, but this has not yet been exercised in a real runtime. See the Checklist.

## What

Adds a complete hermes-agent plugin (`agent-receipts-hermes`) that generates cryptographically signed, hash-linked audit trails for every tool call an agent makes. The plugin classifies tool calls using a bundled taxonomy, forwards frames to the local agent-receipts daemon over AF_UNIX, and exposes two agent-facing tools (`ar_query_receipts`, `ar_verify_chain`) for introspection.

## Why

Provides operators with a tamper-evident audit trail of agent actions. Implements ADR-0010 (Flavor B) where the daemon owns signing, hashing, chain state, and persistence — the plugin&#39;s only job is to classify and forward frames. Mirrors the openclaw plugin&#39;s API so operators can use either implementation with the same muscle memory.

## Implementation

- **Plugin entry** (`__init__.py`): Registers `pre_tool_call` / `post_tool_call` hooks and agent-facing tools via best-effort introspection of the host `ctx`.
- **Hook handlers** (`hooks.py`): Classify tool calls, forward unsigned frames to daemon, correlate pre/post calls via pending map with thread-safe eviction.
- **Classification** (`classify.py`): Tool name → action type + risk level via exact mapping or prefix pattern fallback; supports custom taxonomies that merge with bundled defaults.
- **Daemon integration** (`daemon_store.py`): Read-only access to daemon&#39;s SQLite database; surfaces `DaemonUnavailable` error when daemon is absent.
- **Agent tools** (`tools.py`): `ar_query_receipts` (filter by action/risk/status, return newest-first with stats) and `ar_verify_chain` (verify signatures and hash linkage).
- **CLI** (`cli.py`): Receipt Explorer (`agent-receipts-hermes receipts|verify|export`) for auditing outside agent sessions.
- **Config** (`config.py`): Resolution of daemon socket, database, and public key paths with XDG/env var support.
- **Taxonomy** (`taxonomy.json`): Bundled tool → action mappings and prefix patterns (filesystem, system, browser, network, etc.).

## Testing

- Unit tests for classification, config resolution, tool execution, and hook state management.
- Integration test exercises full plugin lifecycle (register → hooks → real AF_UNIX socket) with a fake daemon server.
- CLI smoke tests verify table and JSON output.
- All tests pass; linter and type checker (strict mode) pass.
- ⚠️ **Not yet manually verified against a real hermes-agent runtime or a real agent-receipts daemon** — see Checklist.

## Security

- Unserialisable tool arguments are dropped from frames, not stringified via `__repr__` (prevents attacker-controllable content in signed audit trail).
- Pending map is guarded by `threading.Lock` to prevent concurrent pre/post invocations from corrupting state.
- No crypto material or chain state held by plugin; daemon is single source of truth.
- Fire-and-forget frame delivery: if daemon socket is unreachable, a one-shot warning is logged and delivery drops silently.

## Checklist

- [x] Tests pass for all changed components
- [x] Linter passes (`ruff check`)
- [x] No real keys or secrets in the diff
- [x] AGENTS.md added with architecture overview and code conventions
- [x] Security: inputs validated at trust boundaries, edge cases tested (nil, empty, concurrent, unserialisable args)
- [ ] **Manual end-to-end verification by the author against a live hermes-agent + agent-receipts daemon — pending; PR stays draft until done**

https://claude.ai/code/session_01Cm5msM2JMUWSzoPnHvJhx2


plugins/hermes/src/agent_receipts_hermes/hooks.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    if not tool_name:
+        return None
+
+    key = _call_key(task_id, session_id, tool_call_id)
+    with state.pending_lock:
+        stashed = state.pending.pop(key, None)
+
+    effective_args = args if args is not None else (stashed.args if stashed else None)
+    classification = classify(tool_name, state.mappings, state.patterns)
```
ojongerius, line 233: **The bundled taxonomy and `taxonomyPath` config have no effect on the actual audit trail.**

The classification machinery here — `classify.py`, `taxonomy.json`, the custom-taxonomy merge/load, and the `taxonomyPath` config option — feeds **only the INFO log line** below (`hooks.py:228-233`):

- `post_tool_call` returns `classification`, but the runtime wrapper `_post` in `__init__.py:197-198` **discards the return value**. Only tests consume it.
- The frame sent to the daemon (`_emit` → `emitter.emit`, `hooks.py:264-272`) carries only `channel, tool_name, decision, input, output, error`. The **daemon does the authoritative classification** from `tool_name`.
- `disclosure_fields` are parsed/validated but **never forwarded anywhere** (and per AGENTS.md, disclosure is governed by the daemon's `--parameter-disclosure` flag, not this plugin).

Consequence: an operator who sets a custom `taxonomyPath` expecting to change receipt action-types/risk gets **no observable effect except log lines**, and there are now two taxonomies (plugin + daemon) that can silently drift. The PR description's lead ("classifies tool calls using a bundled taxonomy, forwards frames") overstates the plugin's role — it forwards `tool_name` and the daemon classifies.

Suggest one of: (a) delete the dead classification path + `taxonomyPath` config + docs (in keeping with "delete unused code ruthlessly"), (b) clearly document it as log-only, or (c) forward the classification to the daemon if the wire protocol can carry it.

---
_Generated by [Claude Code](https://claude.ai/code)_

plugins/hermes/src/agent_receipts_hermes/__init__.py
```diff
@@ -0,0 +1,230 @@
...
+            logger.warning(
+                "agent-receipts: failed to load taxonomy %s: %s — using defaults",
+                cfg.taxonomy_path,
+                exc,
+            )
+
+    emitter: Emitter | None = None
+    if cfg.socket_path:
+        try:
+            emitter = Emitter(socket_path=cfg.socket_path)
```
ojongerius, line 170: **This pins published `0.9.0`, and will invert the "fire-and-forget" guarantee against the monorepo's own in-tree SDK.**

`pyproject.toml` depends on `agent-receipts>=0.9.0` and `uv.lock` resolves the **PyPI wheel**, so tests/CI run against `site-packages/agent_receipts` (0.9.0), not `sdk/py` in this repo. (Contrast the Go modules, deliberately wired via `go.work` so CI catches cross-module breakage.) The in-tree SDK has already moved past 0.9.0 in two breaking ways:

1. **Module/class renamed:** `agent_receipts.emitter.Emitter` → `agent_receipts.daemon_emitter.DaemonEmitter`. The import at `__init__.py:25` (and the type refs in `hooks.py`/tests) won't resolve against in-tree `sdk/py`.
2. **Emit contract inverted (ADR-0025):** in-tree `DaemonEmitter.emit()` now **raises `EmitTransportError` by default** on dial/write failure unless constructed with `best_effort=True`. `EmitTransportError` is a bare `Exception`. This line builds the emitter **without** `best_effort=True`, and `_emit` (`hooks.py:273-280`) only catches `ValueError`/`RuntimeError`.

I confirmed 0.9.0's `emit()` genuinely swallows transport failures, so the "drops silently" claim and `test_emitter_drops_silently_when_socket_unreachable` are **correct today**. But the moment the pin moves past 0.9.0, every tool call will raise into the hermes runtime when the daemon is down — the opposite of the intended safety property — and that integration test will fail.

Suggest: pass `best_effort=True` here and/or defensively catch `EmitTransportError` in `_emit`, and make a conscious decision about whether this plugin should track the in-tree SDK (and if so, wire it so CI catches the divergence).

---
_Generated by [Claude Code](https://claude.ai/code)_

ojongerius (commented):
Two latent issues from a thorough review — neither blocks today's CI (the suite, ruff, and pyright are green against the pinned `agent-receipts==0.9.0`), but both are worth addressing before merge. Inline comments have the detail:

1. **Bundled taxonomy / `taxonomyPath` is log-only** — it doesn't affect the signed audit trail; the daemon does the authoritative classification.
2. **Pinning published `0.9.0` diverges from the in-tree SDK** — bumping past it breaks the import and inverts the "fire-and-forget" guarantee (emit now raises `EmitTransportError` by default under ADR-0025).

The security handling (no-`repr` JSON, adversarial-`__repr__` on-wire test, pending-map locking + stale eviction, read-only store) is solid.

---
_Generated by [Claude Code](https://claude.ai/code)_

plugins/hermes/src/agent_receipts_hermes/classify.py
```diff
@@ -0,0 +1,265 @@
...
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from dataclasses import dataclass
+from importlib.resources import files
+from pathlib import Path
+from typing import Any, cast
```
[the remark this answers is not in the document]
ojongerius: This one is a false positive, so leaving it as-is. `Any` *is* used — inside the quoted `cast(...)` annotations (`cast("dict[str, Any]", …)`, `cast("list[Any]", …)`). Removing the import breaks both linters: `ruff check` reports 6 errors (undefined name in the cast strings) and `pyright --strict` reports 32. The string casts are required because `from __future__ import annotations` defers evaluation, and the names inside `cast()`'s string form still resolve against module scope.

---
_Generated by [Claude Code](https://claude.ai/code)_

plugins/hermes/src/agent_receipts_hermes/hooks.py
ojongerius: Addressed in 7d27f40 by **documenting it as log-only** (your call). The diagnostic-only nature is now stated in `classify.py`'s module docstring, an inline comment at the `classify()` call in `hooks.py`, the README "How it works" + Taxonomy sections, the `taxonomyPath` config row, and `AGENTS.md`. No behaviour change — the daemon remains the authoritative classifier; this just stops the docs implying the plugin's taxonomy reaches the signed receipt.

---
_Generated by [Claude Code](https://claude.ai/code)_

plugins/hermes/src/agent_receipts_hermes/__init__.py
ojongerius: Addressed in 7d27f40, but **not** by adding `best_effort=True` — the pinned `0.9.0` `Emitter.__init__` doesn't accept that kwarg (and has no `EmitTransportError`), so passing it would break against the SDK actually shipped. Instead `_emit` now has a final `except Exception` backstop (with a comment explaining the ADR-0025 contract flip), so transport-class failures are swallowed across SDK versions without importing a type that may be absent. New regression test: `test_emit_transport_error_does_not_propagate`.

The broader "track the in-tree SDK / wire it like `go.work`" decision I've left for you — `sdk/py` is at `0.12.0a1` with the `Emitter`→`DaemonEmitter` rename, but that API isn't released yet, so this stays on `0.9.0` for now and the backstop keeps the forward path safe.

---
_Generated by [Claude Code](https://claude.ai/code)_

plugins/hermes/src/agent_receipts_hermes/classify.py
```diff
@@ -0,0 +1,271 @@
...
```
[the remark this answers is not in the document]
ojongerius: Same false positive as the earlier `Any` thread (#discussion_r3378142620) — re-raised after the push. `Any` *is* used, via the quoted `cast(...)` annotations.

This thread's alternative fix (drop `cast`, write `parsed_dict: dict[str, Any] = parsed`) doesn't work either — I tried it: `cast()` is used in ~5 other spots in this file, so removing it from the import breaks them (**5 ruff errors**), and after the `isinstance(parsed, dict)` guard pyright narrows to `dict[Unknown, Unknown]`, so a bare annotation reintroduces the strict `reportUnknownMemberType` / `reportUnknownVariableType` errors the cast suppresses (**24 pyright errors**). Leaving the `cast` pattern as-is.

---
_Generated by [Claude Code](https://claude.ai/code)_

plugins/hermes/src/agent_receipts_hermes/classify.py
```diff
@@ -0,0 +1,271 @@
...
```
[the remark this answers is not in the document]
ojongerius: Third time — now with the root cause. This file uses the **quoted** `cast("dict[str, Any]", …)` form because the project's own ruff config requires it: rule **TC006** ("Add quotes to type expression in `typing.cast()`").

I tried this exact suggestion (`cast(dict[str, Any], …)`): pyright and pytest stay green, but ruff then fails with **6× TC006**. So there is no form that satisfies both ruff (mandates quotes) *and* this analyzer (won't resolve names inside the quotes) — `Any` is genuinely used. Resolving as a confirmed false positive. The durable fix is analyzer-side: suppress unused-import for names that appear only inside string-form `cast()` arguments.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #802, opened 2026-06-12
fix(daemon): doctor auto-detects the daemon's active chain

ojongerius:
## Problem

`obsigna doctor` run with no `--chain-id` defaulted the chain it inspects to **today's UTC date**:

```go
chainID := fs.String("chain-id", envOr("AGENTRECEIPTS_CHAIN_ID", time.Now().UTC().Format("2006-01-02")), ...)
```

But the daemon does not name its chain after today's date. Its chain id is pinned at startup — from `daemon.toml`'s `chain_id`, the per-OS default, or `AGENTRECEIPTS_CHAIN_ID` — and rolls a `-N` suffix across restarts (`advanceChainID`) when the prior chain is sealed. So on a healthy machine where receipts are actively flowing into, say, `2026-06-03-8`, doctor queried `2026-06-12` and reported:

```
[warn] chain head    chain "2026-06-12" has no receipts yet; nothing to verify
[fail] round-trip    synthetic event was accepted by the daemon but did not land in …/receipts.db
                     within 3s; the daemon may not be the sole writer of this DB
```

The synthetic event **was** traversing the full pipeline and landing correctly — in the live chain. doctor was polling the wrong chain, and the "may not be the sole writer" reason is a misdiagnosis. This also bites any daemon left running across a UTC-midnight rollover.

## Fix

When the operator does not name a chain explicitly (flag, then env), resolve it from the store's **most recently written root chain** instead of the wall clock. The bare UTC date remains only as the fallback for an empty store.

- New `store.LatestRootChainID()` — `SELECT chain_id … WHERE chain_id NOT LIKE '%/agent/%' ORDER BY rowid DESC LIMIT 1`. Recency is by `rowid` (write order), **not** timestamp: a busy daemon writes many receipts within the same RFC3339 second, and `QueryReceipts`' `timestamp DESC, sequence DESC` tiebreak could otherwise surface an older chain's high-sequence receipt over a newer chain's low-sequence one. Agent sub-chains are excluded because the round-trip event and the daemon's primary pipeline write to the root chain.
- `doctor` gains `resolveChainID` / `detectActiveRootChain` (read-only store access — safe against the live daemon as sole writer).

## Drive-by

Doctor's header and error strings now read `obsigna doctor` instead of `agent-receipts doctor`, matching the renamed CLI. (The `agent-receipts-doctor.roundtrip` tool name / `doctor.agent-receipts-doctor.roundtrip` action type are protocol-bound taxonomy constants and were left untouched.)

## Verification

Against a live daemon, no-arg `doctor` now auto-detects the active chain and passes:

```
obsigna doctor — pipeline health
[ok  ] round-trip   synthetic event traversed the full pipeline and landed at seq 340 with a fresh peer credential (pid=80794, uid=501)
doctor: PASS (7 ok, 1 warn, 0 fail)
```

(The remaining warn is the expected "head is not a clean terminator" for a live, open chain.)

## Tests

- `store`: `LatestRootChainID` — empty store, most-recently-written wins over higher-sequence older chain, agent sub-chains excluded.
- `doctorcli`: `detectActiveRootChain` (empty / unreadable / newest root / skips agent sub-chain) and `resolveChainID` (detected wins, empty → date fallback).
- `go test ./...` passes for both `sdk/go` and `daemon`; `go vet` clean.

### DIFF
--- a/daemon/CHANGELOG.md
+++ b/daemon/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`obsigna doctor` auto-detects the daemon's active chain** — with no `--chain-id`/`AGENTRECEIPTS_CHAIN_ID`, doctor now resolves the chain to inspect from the store's most recently written root chain (via `store.LatestRootChainID`) instead of assuming today's UTC date. The old default produced a spurious `chain head` warn and a `round-trip` **fail** ("did not land … the daemon may not be the sole writer") whenever the daemon's chain id was not literally today's UTC date — i.e. a configured `chain_id`, a daemon running across a UTC-midnight rollover, or a rolled `-N` suffix. The synthetic event was traversing the pipeline correctly all along; doctor was just polling the wrong chain. The bare UTC date remains the fallback only for an empty store.
+
+### Changed
+
+- **Doctor output and error messages now say `obsigna doctor`** (not `agent-receipts doctor`), matching the renamed CLI.
+
 ## [0.23.0] - 2026-06-12
 
 ### Changed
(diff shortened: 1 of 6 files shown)


ojongerius:
Follow-up: #803 tracks the remaining `agent-receipts` identifiers the rebrand deliberately left in place (data dir `~/.local/share/agent-receipts/`, `agentreceipts` OS user, `AGENTRECEIPTS_*` env vars, `did:agent-receipts-daemon:` issuer) — out of scope here, surfaced while debugging this doctor false-failure.

PR #807, opened 2026-06-12
feat(collector): binary topology — minimal obsigna-collector, obsigna collector run (ADR-0035)

ojongerius:
## Summary

Applies the binary-topology pattern (ADR-0031 daemon, ADR-0033 mcp-proxy) to the **collector**, one-for-one. Renames the binary `collector` → `obsigna-collector`, makes it the primary minimal entrypoint, and launches it in production via `obsigna collector run` (`syscall.Exec` through the existing generic launcher). The legacy `collector` binary becomes a thin deprecation shim that exec-forwards to `obsigna-collector`.

New **ADR-0035** captures the decision. The collector stays on its own `collector/v*` train and `collector` Homebrew formula for now — ADR-0034 PR-2 folds it into the unified `obsigna` train later, so this PR does **not** pre-consolidate.

## What changed

**Binary topology** (`collector/`)
- `cmd/collector` → `cmd/obsigna-collector` (primary minimal binary, full flag surface).
- New `cmd/collector` deprecation shim: prints a one-line stderr notice and `syscall.Exec`s `obsigna-collector`, forwarding argv + env unchanged (+ tests).
- `.goreleaser.yaml`: two builds (`obsigna-collector` + `collector` shim) with determinism flags (`CGO_ENABLED=0`, `-trimpath`, `-buildvcs=false`, `mod_timestamp: {{ .CommitTimestamp }}`); single `collector` formula installs both binaries with a rename caveat.
- `go.mod`: pin `toolchain go1.26.1` (consumed via `go-version-file`).

**Launcher** (`daemon/cmd/obsigna/`)
- One row added to the launcher table: `collector` → `obsigna-collector`. `TestLauncherSurface` updated. No new launcher code; reuses the generic `syscall.Exec` launcher.

**Gate A — dumb-sink import graph** (`cmd/obsigna-collector/import_guard_test.go`)
- A **structural denylist** (the deliberate inverse of mcp-proxy's fail-closed allowlist): fails on any reach into the **daemon library** (the signer, ADR-0010) or any operator read-side **`*cli`** package. The collector *legitimately* persists receipts (`sdk/go/store`, `sdk/go/receipt`, `modernc.org/sqlite`) — those stay allowed — so the property enforced is "never the signer, never the operator CLI", not "no store".

**Gate B — reproducible build**
- Shared `collector/scripts/reproducible-build.sh`; `collector.yml` PR cross-path double-build byte-identity check; `release-collector.yml` independent rebuild-attest that matches the published archive and publishes the known-good `sha256`.
- `entrypoint_guard_test.go` asserts `obsigna-collector` is the primary build and `collector` appears only as the shim.

**Docs**
- New `docs/adr/0035-collector-binary-topology.md` + index row; `collector/CHANGELOG.md` (rename called out as breaking for installers); `collector/README.md`, `sdk/py/README.md`, `sdk/go/README.md` binary-name updates.

## Gate A: denylist, not allowlist

The collector is a receipt hub (ADR-0017/0020) whose whole job is to persist signed receipts — the opposite of mcp-proxy's thin emitter. An allowlist exists to keep persistence *out*; here persistence is the feature, and enumerating SQLite's large transitive tree would be brittle. So Gate A forbids exactly what the hub must never grow into — the signer and the operator CLI — keyed on the same structural signals as the daemon's Gate A.

## Verification

- `cd collector && go vet ./... && go build ./cmd/... && go test -race ./...` ✅
- `daemon/cmd/obsigna` surface/launcher tests ✅
- Gate A test + cross-path reproducible build (script) verified locally; all workflow YAML parses.

## Out of scope

ADR-0034's unified-train consolidation (PR-2), the Homebrew formula-name migration (no `collector-alpha` track introduced), and the daemon/mcp-proxy/hook restructures. Go module paths and `did:` issuer strings are untouched.

### DIFF
--- a/.github/workflows/collector.yml
+++ b/.github/workflows/collector.yml
@@ -26,16 +26,79 @@ jobs:
 
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
         with:
-          go-version: "1.26"
+          # Pin to the patch recorded in collector/go.mod (ADR-0035) so CI stops
+          # floating to the latest 1.26.x — a moving toolchain would defeat the
+          # reproducible-build attestation Gate B asserts.
+          go-version-file: collector/go.mod
 
       - name: Vet
         run: go vet ./...
 
       - name: Build
-        run: go build ./cmd/collector
+        # -trimpath mirrors the release build (collector/.goreleaser.yaml): CI
+        # must exercise the same flags the attestation depends on (ADR-0035).
+        # ./cmd/... covers both the obsigna-collector binary and the collector shim.
+        run: go build -trimpath ./cmd/...
 
       # -race because the in-memory and SQLite stores carry concurrency
       # invariants exercised by TestInMemoryStore_ConcurrentInserts and
       # TestSQLiteStore_ConcurrentInserts.
       - name: Test
         run: go test -race ./... -v
+
+  # Gate A (ADR-0035): obsigna-collector stays a dumb append-only sink — its
+  # import graph must never reach the daemon library (the signer, ADR-0010) or
+  # any operator read-side (*cli) package. Persistence (sdk/go/store, a SQLite
+  # driver) is the hub's legitimate job and is allowed, which is why this is a
+  # denylist rather than the allowlist obsigna-mcp uses. The rule lives in a Go
+  # test next to the code (cmd/obsigna-collector/import_guard_test.go) so it is
+  # enforced locally too; this job just runs it.
+  import-graph:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
+        with:
+          go-version-file: collector/go.mod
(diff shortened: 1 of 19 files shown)


ojongerius:
Thanks @Copilot — triaged the three comments:

- **`*cli` regex vs. comment (import_guard_test.go)** — fixed in d0fa753. Anchored the rule on an `internal/` segment so it matches the documented `internal/*cli` target exactly and can't surprise-deny an unrelated package whose path merely ends in `cli`. Gate A still passes.
- **Capture `go list` stderr** — not taken here, deliberately. This Gate A test is a near-verbatim mirror of `mcp-proxy/cmd/obsigna-mcp` and `daemon/cmd/obsigna-daemon`, which both use plain `.Output()`. Changing only collector would diverge the three guards; if we want richer failure output it should land across all three as one change.
- **`//go:build darwin || linux` tags** — not taken, same reason. `obsigna-daemon`, `obsigna-mcp`, and their shims all use `syscall.Exec`/`syscall.SIGTERM` with **no** build tags; the darwin/linux restriction is enforced at release (goreleaser `goos:`) and documented in ADR-0031/0033/0035. Adding tags to collector alone breaks the mirror; it's a repo-wide decision if we want it.

Issue #719, opened 2026-06-02
Dashboard reports a valid chain as broken (chain-verification false negative)


ojongerius:
Found while running the dashboard getting-started journey end-to-end (doc e2e fleet, "Priya" persona).

## Summary
The dashboard's chain verification reports a **valid** chain as **broken** — a false negative. This defeats the documented capability ("Chain verification — validates hash linkage", `site/src/content/docs/dashboard/overview.mdx`) and would mislead an operator trying to sanity-check a database.

## Evidence
On a fresh 3-receipt `default` chain:

- `agent-receipts verify` → `Chain default: VALID (3 receipts)`
- Independent check of `/api/receipts` confirms the links are intact:
  - `seq2.previous_receipt_hash == seq1.receipt_hash` ✅
  - `seq3.previous_receipt_hash == seq2.receipt_hash` ✅
- Yet `GET /api/chains/default/verify` returns:

```json
{"valid":false,"broken_at":1,"receipts":[
  {"index":0,"hash_link_valid":true},
  {"index":1,"hash_link_valid":false},
  {"index":2,"hash_link_valid":false}
]}
```

## Root cause
The dashboard recomputes the SHA-256 canonical-form digest **differently from the SDK**, so its recomputed hash doesn't match the stored `receipt_hash` and it flags a false break. This lives in the **dashboard repo** (`agent-receipts/dashboard`, observed at **v0.3.0**), not in this repo — filing here because it's where the affected docs live and this repo is in scope for the fleet.

## Suggested fix
Align the dashboard's hash recomputation with the SDK's RFC 8785 / proof-excluded canonicalisation so it agrees with `agent-receipts verify`.

Related (same persona run):
- Misleading "Verify signatures" button — the endpoint does no signature verification (only `hash_link_valid` / `sequence_valid`). Docs disclaimer added in the accompanying PR; the UI relabel belongs in the dashboard repo.

## Docs mitigation (separate PR)
Until the code is fixed, `dashboard/overview.mdx` now (a) states the dashboard does **not** verify Ed25519 signatures, and (b) warns that chain verification may report false negatives and to confirm with `agent-receipts verify`.

ojongerius:
[...]

PR #825, opened 2026-06-13
docs(dashboard): stop pinning the false-negative caveat to v0.3.0

ojongerius:
## What

The dashboard `overview.mdx` "Known issue: false negatives" caution hard-coded `(v0.3.0)`. The shipped dashboard is now v0.8.0, so the pin is stale and tells a reader nothing about whether the caveat applies to their build.

Surfaced by the doc-e2e *Priya/dashboard* persona (#708 harness).

## Change

- Drop the `(v0.3.0)` version pin; reference [#719](https://github.com/agent-receipts/ar/issues/719) (still **open**) as the live source of truth for current status.

The caveat itself stays — #719 is open and the false negative is unconfirmed-fixed on later builds.


ojongerius:
Superseded by the dashboard content pass: that PR removes the #719 caveat entirely (the false negative was fixed in dashboard v0.4.0 and ar#719 is now closed), rather than just de-pinning it from v0.3.0.

PR #857, opened 2026-06-16
feat(demo): obsigna + sbx side-by-side audit demo

ojongerius:
## Summary

- Adds \`demo/\` — a self-contained \`./demo.sh\` that runs an opencode agent inside a Docker sbx microVM and shows two non-overlapping audit layers side by side: \`sbx policy log\` (infrastructure) and \`obsigna verify\` (signed receipt chain)
- Adds \`demo/undici-stub.mjs\` — esbuild stub replacing undici at bundle time (its webidl internals fail in the opencode runtime; HttpEmitter is unused here)
- Also serves as the end-to-end acceptance test for PR #766 (opencode plugin)

## What the demo does

1. Starts \`obsigna-daemon\` on the **host** (signing key never enters the VM)
2. Bridges the daemon's Unix socket to TCP via \`socat\` (macOS host sockets can't be connected from inside Linux containers via bind-mount)
3. Creates an sbx sandbox; inside the container a second \`socat\` creates a Linux Unix socket tunnelled back to the host daemon
4. Runs \`opencode\` inside sbx with a local ollama model. The obsigna plugin (PR #766) emits one signed receipt per tool call
5. Agent writes \`fibonacci.py\`, runs it, attempts an outbound \`curl\` that sbx blocks
6. Prints \`sbx policy log\` + \`obsigna receipt list\` + \`obsigna verify\` side by side

## Notable build fixes for the plugin bundle

- **node:sqlite**: strip store exports from SDK index before bundling — only \`DaemonEmitter\` is needed, not the SQLite-backed store
- **undici**: stubbed out at esbuild time — its webidl internals fail in the opencode runtime
- **num_ctx**: \`devstral-small-2\` hits ollama's 4096-token default with opencode's system prompt; demo auto-creates a \`devstral-demo\` modelfile with \`num_ctx=32768\`
- **socat tunnel**: macOS host Unix sockets return ECONNREFUSED from inside a Linux container — two-hop socat tunnel keeps the daemon on the host

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -1,0 +1,90 @@
+# obsigna + sbx demo
+
+Two non-overlapping audit layers for an AI agent running inside a Docker sbx microVM:
+
+| Layer | Tool | Question answered |
+|-------|------|-------------------|
+| Infrastructure | `sbx policy log` | What did the VM's network policy allow or block? |
+| Agent actions | `obsigna verify` | What did the agent actually do, in what order, with what inputs? |
+
+Neither log alone tells the full story. sbx sees network packets — not tool semantics. obsigna sees signed receipts for every tool call — not network-level verdicts.
+
+## Running the demo
+
+```
+./demo.sh
+```
+
+Prerequisites: `obsigna-daemon`, `obsigna`, `sbx` (authenticated with `sbx login`), `ollama` running locally with `devstral-small-2:latest` pulled.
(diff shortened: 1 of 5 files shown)


ojongerius:
Closed in favour of demo in the new https://github.com/agent-receipts/obsigna-examples repo

